PR #596, opened 2015-01-06
id-prefix bugfix

comerc:


comerc:
would be a shame if my work will be lost 😄

aldeed:
I manually merged your changes into the devel branch. Thanks!

PR #604, opened 2015-01-08
Feature/subforms

aramk:
Initial support for using subforms. See discussion here: https://github.com/aldeed/meteor-autoform/issues/332


fusepilot:
I just ran across a problem with this pull where it seems to be breaking field array functionallity.

I have a [String] type field in a schema.

````
tagIds:
  type: [String]
  optional: true
````

In the update form, the add and remove buttons work properly when the document is saved with ````aldeed:master````. With ````aramk:feature/subforms````, the array shows the invalidated state after saving in the form even though they have been saved correctly in the document.

I made a repro [here](https://github.com/fusepilot/autoform-array-bug).

I don't know autoform well enough to know if this is related to the issue you mentioned with ````AutoForm.resetForm()````. Apologies if it is.

**Comment by aramk** [comment]
Cheers for that feedback @fusepilot. I'm not sure I'll have a chance to look into this just yet.

File: autoform-inputs.js [hunk 1]
```diff
@@ -325,8 +325,15 @@ updateAllTrackedFieldValues = function updateAllTrackedFieldValues(formId) {
 
 getAllFieldsInForm = function getAllFieldsInForm(template) {
   // Get all elements with `data-schema-key` attribute, unless disabled
-  return template.$("[data-schema-key]").not("[disabled]");
+  // return template.$("[data-schema-key]").not("[disabled]");
   // Exclude fields in sub-forms, since they will belong to a different AutoForm and schema.
   // TODO need some selector/filter that actually works correctly for excluding subforms
-  // return template.$('[data-schema-key]').not("[disabled]").not(template.$('form form [data-schema-key]'));
+  return template.$('[data-schema-key]').not("[disabled]").not(template.$('form form [data-schema-key]'));
```
aldeed: I'm pretty sure this selector won't select any fields in a subform. That's why I had commented it out. Because in a subform `template.$('form form [data-schema-key]')` will match every input.

aramk: I think that's fine though if getAllFieldsInForm() should only select the fields for that form and not the subform? Then the subform will perform this same process and only select it's own direct inputs.

aldeed: No because when called for the subform, `template.$('form form [data-schema-key]')` will match every subform input. Because in the DOM they all have two form element ancestors.

aldeed: Actually, maybe you're correct. I thought only the `[data-schema-key]` was scoped to the template DOM, but it seems like maybe both form ancestors are also scoped to the template.

PR #623, opened 2015-01-12
API for access to internal formData

fusepilot:
I needed this to get a form’s SimpleSchema. Specifically, I needed to specify the SS for a call to RemoveFromFieldAtIndex as it requires it. But I think this will be useful in other scenarios as well.


aldeed:
This and other similar functions have been added to the API in devel branch (5.0).

Issue #383, opened 2014-10-12
text doubles with contenteditable


benjamin79:
Hi,

when using   {{> afFieldInput type="contenteditable"  name="editor" class="form-control" }} the content duplicates on submit. I know there is a bug with meteor / blaze and a possible solution:
https://github.com/Swavek/contenteditable but i don´t know how to connect it with autoform. Please help.
thanks

comerc:
"reactive: false" is works as workaround, thanks.

comerc:
"reactive: false" is works, thanks.

aldeed:
Done in AutoForm 5.0, now released.

PR #673, opened 2015-02-03
text doubles with contenteditable - issue #383

MichalW:
My proof of concept to fix issue #383  (text doubles with contenteditable)

### DIFF
--- a/inputTypes/contenteditable/contenteditable.html
+++ b/inputTypes/contenteditable/contenteditable.html
@@ -1,3 +1,4 @@
 <template name="afContenteditable">
-  <div contenteditable="true" {{this.atts}}>{{{this.value}}}</div>
+  {{getValue this.value}}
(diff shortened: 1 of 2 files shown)


aldeed:
@MichalW I'm having trouble reproducing this issue anymore, so it's tough to see if this fixes it. Do you have a reliable way to reproduce? I'm testing using `devel` branch.

MichalW:
@aldeed The problem still exists, also on "devel". It is very random - sometimes work, sometimes not.
I will try to find reliable way to reproduce.

MichalW:
I created small project to reproduce problem:
https://github.com/MichalW/meteor-contenteditable-test

I recorded what you need to do to make the problem appear:
https://www.youtube.com/watch?v=rZuwS454fDs&feature=youtu.be

aldeed:
OK, seems to work and fix the issue as far as I can tell. Thanks for the PR and good repro!

Issue #513, opened 2014-12-08
Feature Request: autosave on throttled keyup


comerc:
Like autosave function. When I entered some data, then Meteor saved it on timer.

aldeed:
Done in dev branch with PR #513

aldeed:
Done in AutoForm 5.0, now released.

PR #685, opened 2015-02-11
throttled autosaveOnKeyup; #513

jfols:
It would be nice to add an option to set the throttle time, but I'm not sure where to pull that option from the way I'm creating the throttled autosave function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,8 @@ Meteor method in this attribute.
 * `resetOnSuccess`: Optional. The form is automatically reset
 for you after a successful submission action. You can skip this by setting this
 attribute to `false`.
-* `autosave`: Optional. Set to `true` to enable automatic form submission for a `type="update"` form. Whenever any field value changes within the form, the change will be automatically saved to the database.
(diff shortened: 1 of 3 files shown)


jfols:
I created the `autosaveOnKeyup` option to differentiate from the existing `autosave` option. It's quite likely that some applications will want the current behavior of `autosave` while others may want `autosaveOnKeyup`.

aldeed:
In the `change form` event, I think it should also autosave when `autosaveOnKeyup` is `true`, as a safety. I recall issues in the past where keyup doesn't necessary have the final correct value, so best to save it on change, too.

Also can you change the two instances of the word "thrown" to "emitted" in the readme text?

Otherwise looks mergeable. Thanks for doing it.

jfols:
Done and done :smile: No problem man, thanks for all you work on this package!

jfols:
Should probably recommend `trimStrings=false` option in conjunction with `autosaveOnKeyup`, otherwise spaces may be deleted while typing.

aldeed:
Good point re `trimStrings`. I'll add that to the readme.

PR #698, opened 2015-02-17
fixed button align in bootstrap3-inline by increasing the specificity of...

nimerritt:
When using the bootstrap3-inline template the button was aligned to middle and not to the top. Increased the specificity of the autoform-align-inline selector to override the vertical-align: top, which comes from bootstrap's .btn class.


nimerritt:
image

aldeed:
Can you issue PR to `devel` branch instead? Some of the styling has already been tweaked and fixed in that branch, though I think this particular fix is probably still needed.

nimerritt:
Done, see PR #700

PR #710, opened 2015-02-21
Forms states

mxab:
Hi I started working on the form dirty/clean/pristine 
not sure if it is the right way to do it, but the example seems to work https://github.com/mxab/autoform-states-sample

### DIFF
--- a/autoform-api.js
+++ b/autoform-api.js
@@ -527,3 +527,19 @@ AutoForm.invalidateFormContext = function autoFormInvalidateFormContext(formId) 
   formDeps[formId] = formDeps[formId] || new Tracker.Dependency();
   formDeps[formId].changed();
 };
+
+
+
+AutoForm.getFormState = function autoFormGetFormState(formId) {
+  formStates[formId] = formStates[formId] || new ReactiveVar("pristine");
+  return formStates[formId].get();
+};
+AutoForm.isFormClean = function autoFormGetFormStateClean(formId) {
+  return AutoForm.getFormState(formId) === "clean";
+};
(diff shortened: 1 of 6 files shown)


jcheroske:
A couple of questions. Does your approach allow for fields to also have dirty/pristine state? And, what events cause the various input fields to become dirty?

mxab:
hi, so far it is only the complete form. I use the getFormValues and the form doc and compare it via EJSON

mxab:
After further testing I think to check the clean/dirty state of a document to update it to check if there are any entries in the modifier's $unset and $set.
any thoughts on that @aldeed

mxab:
I did some rework on the "dirty" checking part, seems to work better now

aldeed:
Great, thanks. Might be a few days before I can take a good look at this.

jcheroske:
A couple of cases to consider:

* Does submitting the form make it dirty if it fails validation?

* Does a blur event make the form dirty?

mxab:
@jcheroske at the moment input and change events make it dirty/clean but could at blur easily. At the moment I would not use it, for some reason I had some strange issues in some forms...

chmac:
Any progress on this?

@mxab Up for fixing the merge conflicts?

@aldeed Is it ready to merge (assuming conflicts are resolved)? Any issues with it?

AutoForm is awesome. We're using several `quickForm`s to easily render huge input forms. I'd like to warn the users that they need to save as soon as the form becomes "dirty" and this PR looks like it would be a huge help to do that.

mxab:
Sorry I'm not maintaining this PR anymore, something introduced horrible performance when I used the dirty checking

PR #719, opened 2015-03-01
IE8 compatibility

alethes:
Unfortunately, IE8 still accounts for a couple percent of Internet users. Using the dot notation to access the `for` attribute of an object causes an exception in that browser, and there's an easy way to avoid this.

By the way, thanks for this great package!


aldeed:
Thanks, manually merged in devel branch.

PR #756, opened 2015-03-11
before no addHooks is not documented

aj07mm:
before no addHooks is not documented


aldeed:
I rearranged the text to try to clarify this in a different way. No need to duplicate the explanation of the hooks object since it's the same for both `hooks` and `addHooks` API. Thanks for pointing this out.

aj07mm:
nice!

PR #782, opened 2015-03-16
Fixed AutoForm.Utility.dateToDateString* methods

### LINKED ISSUE #774: AutoForm.Utility.dateToDateStringUTC doesn't handle years < 1000
See the following behavior of [AutoForm.Utility.dateToDateStringUTC](https://github.com/aldeed/meteor-autoform/blob/master/utility.js#L356):

```javascript
var d = new Date();
d.setFullYear(1);
AutoForm.Utility.dateToDateStringUTC(d)
"1-03-15"
```

When using `input type="date"` combined with `autosave=true`, this can cause a `The specified value '2-03-14' does not conform to the required format, 'yyyy-MM-dd'` message when typing in today's date (3/14/2015), because the first valid date chrome sees is 3/14/2, which causes autoForm to do its thing and break.

Am happy to submit a PR to fix this if interested. The correct thing to do here would be to return "0001-03-15" in the example, which will make chrome happy.

jfly:
AutoForm.Utility.dateToDateStringUTC and AutoForm.Utility.dateToDateString now correctly handle dates with years < 1000. This fixes #774.

I added a TinyTest for this, which was a little tricky because we cannot
control the timezone the tests run in (the behavior of AutoForm.Utility.dateToDateString
is influenced by the timezone). I opted to have the test check what
timezone we're in, and assert appropriately (the 3 interesting cases
are: we're ahead of UTC, we're behind UTC, and we're in UTC). To test
these, I ran firefox with the TZ environment variable set:

```bash
TZ=UTC firefox
>> new Date().getTimezoneOffset()
0

TZ=Europe/Amsterdam firefox
>> new Date().getTimezoneOffset()
-60

TZ=America/Los_Angeles firefox
>> new Date().getTimezoneOffset()
420
```

The tests passed under all 3 instances of firefox.

### DIFF
--- a/package.js
+++ b/package.js
@@ -224,5 +224,6 @@ Package.onUse(function(api) {
 
 Package.onTest(function (api) {
   api.use(['aldeed:autoform', 'tinytest', 'underscore']);
+  api.use('momentjs:moment', 'client');
   api.addFiles(['tests/utility-tests.js', 'tests/autoform-tests.js']);
 });
--- a/tests/utility-tests.js
+++ b/tests/utility-tests.js
@@ -290,6 +290,26 @@ if (Meteor.isClient) {
     });
   });
 
+  var startOfDayUTC = moment.utc("0001-3-14", "YYYY-MM-DD").toDate();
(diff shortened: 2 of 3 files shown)


aldeed:
Thanks @jfly.

PR #792, opened 2015-03-18
New Add-On Package (universe-autoform-select)

MichalW:
https://atmospherejs.com/vazco/universe-autoform-select


aldeed:
Thanks!

PR #797, opened 2015-03-19
adds daterangepicker custom input type

antalakas:
valueOut is a date array (from-to), options not supported yet but moving quickly to cover this


aldeed:
Thanks!

PR #807, opened 2015-03-21
Add return to afOptionsFromSchema helper

abecks:
Hey aldeed,

I noticed that my select options were not showing up using the new ```{{> afQuickField name=this.name options=afOptionsFromSchema}}``` format. 

I managed to trace it down to this helper:
```
Template.registerHelper('afOptionsFromSchema', function afOptionsFromSchema() {
  AutoForm._getOptionsForField(this.name);
});
```

I believe it should be:
```
Template.registerHelper('afOptionsFromSchema', function afOptionsFromSchema() {
  return AutoForm._getOptionsForField(this.name);
});
```

Otherwise the results of _getOptionsForField are not being returned to the template.


aldeed:
Great, thanks!

Issue #829, opened 2015-03-26
onSubmit not called anymore in 5.1


garenor:
I upgraded to 5.1.0 and all of a sudden none of my forms are working anymore...

I have the following setup:

````
{{#autoForm schema="shipSchema.exNames" doc=ship id="updateShipExNamesForm"}}
````

and a formhandler:

````
AutoForm.addHooks('updateShipExNamesForm', {
    onSubmit: function (insertDoc, updateDoc, currentDoc) {
        this.event.preventDefault();
        console.log('before update call exnames');
         
        ...
    
        return false;
    },
});
````

All my update forms are the same way, none are working in 5.1.0. No problem in 5.0.3. Any ideas?

**Comment by abecks** [comment]
Are you getting validation errors? I'm now getting:

```
SimpleSchema invalid keys for "stepViewForm" context:
[Object, Object]
0: Objectname: "insertDoc"type: "keyNotInSchema"value: Object__proto__: Object
1: Objectname: "updateDoc"type: "keyNotInSchema"value: Object__proto__: Objectlength: 2__proto__: Array[0]
```

It's trying to validate against insertDoc and updateDoc, instead of the contents of either. I'll see if I can track it down but probably need @aldeed for this one.

**Comment by garenor** [comment]
No, I also do not get any validation errors on most forms. It's just plain not working anymore. I only get a validation message on one form, while the provided input in the formfield is ok (works ok with 5.0.3).

**Comment by abecks** [comment]
Interesting, I'm getting validation errors for forms that do not have a type set.

Can you paste this into your project somewhere and see if they start working (only ones without a type set should be affected):

```
AutoForm.addFormType('normal', {
  onSubmit: function () {
    var c = this;

    // Get onSubmit hooks
    // These are called differently from the before hooks because
    // they run async, but they can run in parallel and we need the
    // result of all of them immediately because they can return
    // false to stop normal form submission.
    var hooks = Hooks.getHooks(this.formId, 'onSubmit');

    var hookCount = hooks.length, doneCount = 0, submitError, submitResult;

    if (hookCount === 0) {
      // we haven't called preventDefault, so normal browser
      // submission will now happen
      this.endSubmission();
      return;
    }

    // Set up onSubmit hook context
    var ctx = _.extend({
      done: function (error, result) {
        doneCount++;
        if (!submitError && error) {
          submitError = error;
        }
        if (!submitResult && result) {
          submitResult = result;
        }
        if (doneCount === hookCount) {
[...]

PR #830, opened 2015-03-26
Fix validation for autoForm's with no type set

abecks:
Hey @aldeed, as per issue #829, forms without a type are no longer working. I was getting validation errors in console due to:

```
return AutoForm._validateFormDoc(this.formDoc, false, this.form.id, ss, this.form);
```

where formDoc contains both ```insertDoc``` and ```updateDoc```

I believe changing this to:

```
return AutoForm._validateFormDoc(this.formDoc.insertDoc, false, this.form.id, ss, this.form);
```

will fix the issue. I can't imagine forms without a type wanting to use the updateDoc for validation anyway.

### DIFF
--- a/formTypes/normal.js
+++ b/formTypes/normal.js
@@ -59,6 +59,6 @@ AutoForm.addFormType('normal', {
     // Get SimpleSchema
     var ss = AutoForm.getFormSchema(this.form.id);
     // Validate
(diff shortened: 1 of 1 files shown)


aldeed:
Thanks, @abecks. I appreciate the debugging. This is the correct fix. Unlike other forms, we generate both insert doc and update modifier for typeless forms because we don't know whether the onSubmit will be doing an insert or an update. But for validation, yes, we use the insert doc.

Issue #860, opened 2015-04-05
Fieldset Support


SachaG:
So far I've used my own AutoForm template in Telescope to add fieldsets that let me group fields together inside a single Quickform. 

But having my own template with its own custom logic isn't really maintainable, so I'd like to try and submit a PR to add fieldsets to AutoForm core, if that's a feature that makes sense.

Any tips on where I should start?

SachaG:
So I looked into modifying the `afQuickFields` template, but I'm not quite sure how to do it because it doesn't seem like this template has the fields as its data context like I thought it would. So I'm not entirely sure where the field data is coming from, and how to iterate over it to extract the fieldsets…

aldeed:
The [afFieldNames helper](https://github.com/aldeed/meteor-autoform/blob/master/autoform-helpers.js#L119) has most of the magic logic. It returns a list of field names based on a number of factors. If `name` is set, then the fields are limited to fields at and below that field name in the tree. Otherwise all fields in the schema. But then `fields` and `omitFields` attributes are also checked, further limiting that list.

I'm not sure how easy it will be to add your logic there, so it might have to be done at a higher level. Maybe the `quickForm` template can check for fieldsets and call `afQuickFields` multiple times, once for each fieldset, passing in the fieldset name as yet another limiter that `afFieldNames` will respect.

SachaG:
Yes, that's the approach I was going for with `afQuickFields`:

```html
<template name="afQuickFields">

  <div class='no-fieldset'>
    {{#each afFieldNames name=this.name}}
      {{log}}
      {{> afQuickField quickFieldAtts}}
    {{/each}}
  </div>
  
  {{#each afFieldsets}}
      {{log}}
      <fieldset>
        <h3 class="af-fieldset-heading">{{this.fieldsetName}}</h3>
        {{#each afFieldNames name=this.name fields=fieldsForFieldset}}
          {{log}}
          {{> afQuickField quickFieldAtts}}
        {{/each}}
      </fieldset>
  {{/each}}

</template>
```

But I got stuck when trying to generate `afFieldsets` and `fieldsForFieldset`, since I wasn't sure how to access the current schema. Maybe we could Skype about it sometime?

SachaG:
OK, I'm making some progress. Not very far now :)

PR #865, opened 2015-04-07
Split form in fieldsets according to autoform.group option

SachaG:


SachaG:
See #860.

The idea is to figure out which fieldsets are defined based on the schema, thensetting the `fields` property for each fieldset and passing that on to `innerContext`.

This works, except for the fact that it repeats the Submit button… But hopefully we can work from here :)

SachaG:
Apart from changing the template and adding new helpers, the only things I changed were:

Set `atts` to the parent data context instead of `this` (since it's now inside a `with` or `each`):

```
 var atts = Template.parentData(1);
```

Set the `fields` attribute on `atts`:

```
atts.fields = this.fields;
```

SachaG:
Oh also, just in case it's not clear this works based on any `autoform.group` option you've defined in your schema, there's no separate master list of fieldsets. This means it's not really possible to set fieldset order.

Also I'm not sure how to internationalize fieldset labels yet. Maybe we need a small helper for that based on whatever is used for form field labels.

aldeed:
Looks like a good start. The base templates (`quickForm`, etc.) are a layer that does all of the generic stuff, and then they pass that off to the theme templates (`quickForm_plain`, `quickForm_bootstrap3`, etc.) for markup.

So you should:

1. Add `fieldsWithNoFieldsets` and `fieldsets` to the object returned by the innerContext helper, instead of having them be template helpers.
2. Move your logic in the HTML to the theme templates instead:
  * https://github.com/aldeed/meteor-autoform/blob/master/templates/plain/components/quickForm/quickForm.html
  * https://github.com/aldeed/meteor-autoform/blob/master/templates/plain-fieldset/plain-fieldset.html
  * https://github.com/aldeed/meteor-autoform/blob/master/templates/bootstrap3/components/quickForm/quickForm.html

Your same template logic should work in those files, but surrounding the `afQuickFields` instead of the `Template.dynamic`. We might be able to deprecate the "plain-fieldset" template if your changes take care of that use case.

The reason for keeping it in the theme template is so that anyone could make their own custom theme to adjust the way they're shown. Someone might want to use the `group` feature to render panels, accordions, etc. instead of fieldsets.

SachaG:
OK, done for the `bootstrap3-horizontal` template. If that looks good to you I'll do it for the others as well. I also made sure to make it work if the `fields` parameter was already specified, which I wasn't doing before. 

I do wonder if we're not going to end up repeating the logic/markup a lot for each template though?

SachaG:
It turns out that I was running into the old problem that no `fields` means "show everything", not "show nothing". I just added a workaround.

SachaG:
Any idea when you'd be able to look at this? Just trying to figure out the timeline of when I can expect to pull back those changes into Telescope :)

components/quickForm/quickForm.js
```diff
@@ -6,7 +6,45 @@ Template.quickForm.helpers({
...
+
+    // if atts.fields exists, restrict the schema to specified fields
+    if (atts.fields) {
+      // note: haven't found a clean way to do this all in one go yet, so doing it later. 
+    }
+
+    // get list of unique field groups with any falsy values removed
+    // note: if atts.fields is specified, only consider fields contained in it
+    var fieldGroups = _.compact(_.unique(_.map(schema, function (property, key) {
+      return (!atts.fields || _.contains(atts.fields, key)) && property.autoform && property.autoform.group ;
```
aldeed: You will have to do `key = SimpleSchema._makeGeneric(key);` before this because otherwise `_.contains(atts.fields, key)` won't work when the key is something like `emails.0.address`, which is `emails.$.address` in the schema.

components/quickForm/quickForm.js
```diff
@@ -6,7 +6,45 @@ Template.quickForm.helpers({
...
+    }))).sort();
+
+    // build fieldGroups array
+    var fieldGroups = _.map(fieldGroups, function (property, key) {
```
aldeed: Should be no `var` before `fieldGroups` here since it's already defined.

components/quickForm/quickForm.js
```diff
@@ -6,7 +6,45 @@ Template.quickForm.helpers({
...
+      };
+    });
+
+    // get all fields with no field group specified
+    // note: if atts.fields is specified, only consider fields contained in it
+    // always omit fields with "$" in their name
+    var fieldWithNoGroups = _.compact(_.map(schema, function (property, key) {
+      return key.indexOf("$") == "-1" && (!atts.fields || _.contains(atts.fields, key)) && (!property.autoform || !property.autoform.group) && key;
+    }));
+    var fieldWithNoGroups = {fields: fieldWithNoGroups};
```
aldeed: Should be no `var` before `fieldWithNoGroups` here since it's already defined.

templates/bootstrap3-horizontal/components/quickForm/quickForm.js
```diff
@@ -5,10 +5,38 @@ Template['quickForm_bootstrap3-horizontal'].helpers({
...
-    return _.pick(this.atts, 'id-prefix', 'input-col-class', 'label-class');
+
+    // get fields for current field group
+    var fieldsForCurrentGroup = this.fields;
+
+    // clone the object to make sure we don't modify the original data context
+    var atts = _.extend({}, Template.parentData(1).atts);
+    
+    atts.fields = fieldsForCurrentGroup
+    return _.pick(atts, 'id-prefix', 'input-col-class', 'label-class');
```
aldeed: This should include `fields` otherwise you're stripping off the property you just added. I think it may also look back up the data context tree for a `fields` attribute, so it might even work without setting `fields` explicitly here. I don't remember the exact details.

templates/bootstrap3-horizontal/components/quickForm/quickForm.js
```diff
@@ -5,10 +5,38 @@ Template['quickForm_bootstrap3-horizontal'].helpers({
   labelClass: function () {
     return this.atts["label-class"];
   },
+  fieldGroupLabel: function () {
+    var name = this.name;
+
+    // if field group name is of the form XY_abcde where "XY" is a number, remove prefix
+    if (typeof parseInt(name.substr(0,2)) === "number" && name.charAt(2) === "_") {
+      var name = name.substr(3);
```
aldeed: Should be no `var` before `name` here since it's already defined.

templates/bootstrap3-horizontal/components/quickForm/quickForm.js
```diff
@@ -5,10 +5,38 @@ Template['quickForm_bootstrap3-horizontal'].helpers({
   labelClass: function () {
     return this.atts["label-class"];
   },
+  fieldGroupLabel: function () {
+    var name = this.name;
+
+    // if field group name is of the form XY_abcde where "XY" is a number, remove prefix
+    if (typeof parseInt(name.substr(0,2)) === "number" && name.charAt(2) === "_") {
```
aldeed: You should always specify `10` for the second arg of `parseInt` because some browsers use base 2 or some other base.

Also, this check doesn't really work because `typeof parseInt('foo')` is "number", which is because `typeof NaN` is, paradoxically, "number". So instead it should be `if (!isNaN(parseInt(name.substr(0,2), 10)) && name.charAt(2) === "_")`.

aldeed:
@SachaG, other than my inline comments, I have one general concern that I might need to test out to be fully sure what I'm talking about, but essentially I'm just worried that you're circumventing some of the field selection logic. There is a bunch of logic in `afFieldNames` that ensures that all the logic described [here](http://autoform.meteor.com/qfdetails) works properly. For example, if I include `address.city` in `fields`, it should include only that single input form group, but if I include both `address` (an Object type) and `address.city`, then it should make an `afObjectField` for `address` (which is a panel when using bootstrap), and *within that panel*, there should be only the `address.city` form group and none of the others.

There is similar logic for when to display single fields for editing arrays versus a full `afArrayField` with plus/minus buttons.

As long as all those unique cases still work the same, I'm fine with this approach. If those aren't working, then I would need to think about it a bit more to figure out how we can keep that logic while adding yours.

SachaG:
Thanks for the feedback, I first the first couple issues. For the `fields` thing, I'll clone the autoform demo locally and go through each of the examples to compare it with the current live demo.

SachaG:
So there is a problem with the `afFieldNames` helper (`Exception in template helper: TypeError: Cannot read property 'autoform' of undefined`), but I'm a bit lost to be honest… I think I may be passing the wrong data context or argument, but it's hard to pinpoint where. 

Maybe you can take a look later, or we can quickly Skype so you can point out what I'm doing wrong?

SachaG:
OK, so I think I figured out a good part of the previous issues, and managed to simplify the logic as well. 

Basically what I'm doing is:

1. Generate data contexts for each field group, plus one for fields with no groups.
1. Take the `quickForm`'s original `atts` property and clone it to each of these contexts.
1. For each group, take the intersection of the `quickForm`'s `fields` property, and the fields that belong to the group.
1. Pass the whole thing on (original `atts`, `fieldGroups`, `fieldsWithNoGroups`) to the template. 

This seems to render every form in the demo as it should, except for this one where the field order is *not* reversed with my version:

```
{{> quickForm id="fields3" schema=Schemas.FieldsExamples fields="name,contacts,contacts.$.phone,contacts.$.name"}}
```

SachaG:
I wasn't handling getting the schema from `atts` properly, this should work a little better (although I'm still running into errors on the "updatepush" example).

SachaG:
OK, a little more progress :) I simplified the logic to only keep a single `{{#each}}` loop for both fields with and without a group. 

By the way, it'd be cool if templates could inherit helpers from each other. For example, the `fieldGroupClass` and `fieldGroupLabel` helpers are exactly the same for both `bootstrap` and `bootstrap-horizontal`. I'm not sure what the best approach would be to avoid code duplication?

SachaG:
A couple other questions:

- Is there a pre-made function to get an array of field names from the `fields` attribute?
- Is there a pre-made function to get a schema from either the `schema` or `collection` attributes?

SachaG:
OK, so I fixed the field ordering thing. Apparently I wasn't handling `contacts.0.name` and `contacts.$.name` very well, but I think I understand that a bit better now. 

The updatePush issue still remains, I'm not sure how to point it to the correct schema…

SachaG:
Fixed another issue I had overlooked. Any idea when you might be able to look at this again? I would love to get this wrapped up so I can include it in the next version of Telescope, and move on to working on other improvements to AutoForm :)

SachaG:
Woohoo! Did you get a chance to fix the updatePush thing though?

aldeed:
I'm doing some testing before releasing. I'll fix that and whatever I find.

SachaG:
Cool, let me know if I can help :)

fermuch:
Woohoo! I'm loving autoform.group :smiley:

Issue #861, opened 2015-04-06
afFieldValueIs and AutoForm.getFieldValue not reactive with checkboxes


erasaur:
The helper returns true when the checkbox is initially checked, but doesn't return false when the checkbox is subsequently unchecked. Same behavior with `AutoForm.getFieldValue`.

```
<head>
  <title>test-checkbox</title>
</head>

<body>
  {{> hello}}
</body>

<template name="hello">
  {{#autoForm collection="FieldValueIs" type="insert" id="FieldValueIsForm"}}
    {{> afQuickField name="a"}}
    {{#if afFieldValueIs name="a" value=true}}
      {{> afQuickField name="b"}}
    {{/if}}
  {{/autoForm}}
</template>
```

```
FieldValueIs = new Mongo.Collection("FieldValueIs");
FieldValueIs.attachSchema(new SimpleSchema({
  a: {
    type: Boolean
  },
  b: {
    type: String
  }
}));
```

marktomm:
Hello!

I have also stumbled upon this buggy behavior in autoform.

The only workaround I have found so far is set a Session variable on click event if I want to dynamically change properties in the schema:

<pre>
//hello.js

Template.hello.events({
  "click #a": function() {
    Session.set('form_a_Option', AutoForm.getFieldValue('a', 'FieldValueIsForm'));
  }
});


//colleciton.js

FieldValueIs = new Mongo.Collection("FieldValueIs");
FieldValueIs.attachSchema(new SimpleSchema({
  a: {
    type: Boolean
  },
  b: {
    type: String,
    autoform: {
      type: function() {
        if(!Session.get('form_a_Option'))
          return "hidden";
          
        return "";
      }
    }
  }
}));
</pre>

or write a custom helper for hello template

<pre>
Template.hello.helpers({
  is_a_OptionYes: function() {
    return AutoForm.getFieldValue('a');
  }
});
</pre>

The examples here assume an id "a" is assigned to the appropriate field i "hello" template:
{{#if afFieldValueIs name="a" id="a" value=true}}

abecks:
Confirmed, I'll see if I can diagnose and PR a fix.

abecks:
This is fixed in #889, monitor that pull request to see when the package is updated.

rclai:
What crazy timing, I created a [reproduction](https://github.com/rclai/autoform-affieldvalueis-issue) just now lol.

chompomonim:
[...]

PR #889, opened 2015-04-15
Fix for checkboxes and radio buttons not triggering form change event

abecks:
Hey Aldeed,

This is a fix for #861. Checkboxes and radio buttons do not pass the infinite loop protection you have in place on the ```change form``` event handler. This adds compatibility for checkbox and radio buttons, and provides a framework for other input types if necessary in the future.

Cheers!


abecks:
Actually I see I've made a mistake here, going to redo the request.

PR #890, opened 2015-04-15
Fix for checkboxes and radio buttons not triggering form change event

abecks:
Hey Aldeed,

This is a fix for #861. Checkboxes and radio buttons do not pass the infinite loop protection you have in place on the change form event handler. This adds compatibility for checkbox and radio buttons, and provides a framework for other input types if necessary in the future.

Cheers!


rclai:
There is something not working right with nested `afFieldValueIs` even with this fix.

Can you test [this](http://meteorpad.com/pad/pQTK2vM7MjFJ5fbMb/AutoForm%20afFieldValueIs%20Issue) with your fix?

rclai:
In 4.2.2, this would work properly as shown [here](http://meteorpad.com/pad/F3MZPmbYMAw6DKEWR/Copy%20of%20AutoForm%20afFieldValueIs%20Issue). You will notice that text that is supposed to be hidden appear but disappear.

abecks:
Your [example](http://meteorpad.com/pad/pQTK2vM7MjFJ5fbMb/AutoForm%20afFieldValueIs%20Issue) works for me with my fix, however I had to turn your {{#unless}} block into an {{#if}} to show the string properly. It does reset the value of the second checkbox to true when the template is re-rendered due to toggling on the first one. However I do see some complications stemming from the infinite loop protection:

Set the first checkbox to true,
Set the second checkbox to false,
Set the first checkbox to false and then true again

The second checkbox will now be to true, and setting it to false will not trigger a change event (due to the last remembered state of the field being false, it thinks no change has occured).

I'm going to have to defer to @aldeed for any complications due to the loop. I don't know what the original problem was that he was trying to avoid.

If my fix isnt working for you, can I ask how you are using it? You can checkout my version of the repository to ```packages/aldeed_autoform``` in your Meteor directory to override it.

rclai:
I had no solution, until you just mentioned how you inverted the logic check from `unless` to `if` and this time it worked properly, wow.

abecks:
You might still encounter issues due to the re-rendering of the second checkbox without remembering its setting, but that should get you going.

roelvan:
I don't like having to install a custom autoform package and change all related dependencies, so I've built a temporary fix by creating a custom input type for checkboxes: You can find it here: https://github.com/roelvan/autoform-checkbox-fix

jasonxeno:
This pull request is working great for me in production, thank you!

tcastelli:
working for me too, hope to see that integrated soon :)

aldeed:
Thanks! Manually merged in https://github.com/aldeed/meteor-autoform/commit/b2fb325f6021a59d7938a05b075a2593df62b35c

PR #927, opened 2015-05-05
Update the themes section

JamesLefrere:
I'm no longer maintaining that Semantic UI package, so I've pointed the link towards the new one.


aldeed:
Thanks!

PR #939, opened 2015-05-13
Options for afFormGroup and afObjectField

fknipp:
I extended the autoform attributes for “nicer” forms using the Bootstrap grid, e. g. to display first name and surname or ZIP and city side by side.

Example:

```javascript
Schema.Registration = new SimpleSchema({
	'participant': {
		type: Object,
		autoform: { 
			afObjectField: {
				bodyClass: 'container-fluid row'
			}
		}
	},
	'participant.firstname': {
		type: String,
		autoform: { 
			afFormGroup: {
				'formgroup-class': 'col-sm-6'
			},
		}
	},
	'participant.lastname': {
		type: String,
		autoform: { 
			afFormGroup: {
				'formgroup-class': 'col-sm-6'
			},
		}
	},
	…
};
```

github firstname lastname

### DIFF
--- a/components/afObjectField/afObjectField.js
+++ b/components/afObjectField/afObjectField.js
@@ -6,7 +6,6 @@ Template.afObjectField.helpers({
   },
   innerContext: function() {
     var c = AutoForm.Utility.getComponentContext(this, 'afObjectField');
-    _.extend(this, c.atts);
-    return this;
+    return _.extend({{}, this, c.atts);
   }
 });
--- a/components/afObjectField/afObjectField.js
+++ b/components/afObjectField/afObjectField.js
@@ -6,6 +6,6 @@ Template.afObjectField.helpers({
   },
   innerContext: function() {
     var c = AutoForm.Utility.getComponentContext(this, 'afObjectField');
-    return _.extend({{}, this, c.atts);
+    return _.extend({}, this, c.atts);
   }
 });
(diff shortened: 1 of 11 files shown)


components/afObjectField/afObjectField.js
```diff
@@ -3,5 +3,10 @@
 Template.afObjectField.helpers({
   getTemplateName: function () {
     return AutoForm.getTemplateName('afObjectField', this.template, this.name);
+  },
+  innerContext: function() {
+    var c = AutoForm.Utility.getComponentContext(this, 'afObjectField');
+    _.extend(this, c.atts);
+    return this;
```
aldeed: We should never mutate `this`. Maybe do `return _.extend({}, this, c.atts);` instead.

aldeed: Actually, most likely just `return c` will work.

aldeed:
Thanks, @fknipp! This is on my to-do list, and I think you've done a good job with it. I left a couple code comments, but if we can resolve that, this should be good to merge.

components/afObjectField/afObjectField.js
fknipp: Thanks for pointing this out.

I've pushed a version using the first version, at I cannot assume, how this is used in the templates used by other users.

PR #946, opened 2015-05-15
Allow user to specify a DDP Connection to use when calling Meteor methods

patrickleet:
```html
{{#autoForm schema=myForm id="myForm" type="method" meteormethod="doStuff" ddp=DDPConnection}}
  <!-- ... -->
{{/autoForm}}
```

Just pass in DDPConnection from a template helper

```javascript
Template.myForm.helpers({
    DDPConnection: function() {
        return DDP.connect(url); 
        // return Cluster.discoverConnection('service'); // also works with meteor-cluster!
    }
});
```


aldeed:
Manually merged. Thanks!

PR #947, opened 2015-05-18
Remove 'autosave' value from 'removeProps'

MichalW:
I working on "universe-select" package and I need "autosave" value in contextAdjust in "addInputType".
Could You leave "autosave" value in context?


aldeed:
@MichalW, you can do `AutoForm.getCurrentDataForForm()` in the contextAdjust function to get the reactive autoForm attributes.

PR #985, opened 2015-05-30
Reactive schema

abecks:
Hey @aldeed,

I think I've got reactive schemas working with only minor adjustments to your code. It mostly consists of avoiding lookups for fields that no longer exist, and capturing errors thrown by ```Utillity.getDefs```.

What I've found is that some of field template helper's have a reactive dependency on ```Blaze.getData``` from ```Utility.getComponentContext```, and anything that ultimately uses ```AutoForm.getCurrentDataForForm```. When the schema changes, these template helpers re-run, before Blaze actually removes them from the DOM. It seems that we will need checks in place on the helpers (or these utility functions) to check to make sure the field still exists in the schema.

I've tested it in my production app without any issues, but should probably be tested thoroughly.

### DIFF
--- a/autoform-api.js
+++ b/autoform-api.js
@@ -477,9 +477,9 @@ AutoForm.getInputTypeTemplateNameForElement = function autoFormGetInputTypeTempl
 AutoForm.getInputValue = function autoFormGetInputValue(element, ss) {
   var field, fieldName, fieldType, arrayItemFieldType, val, typeDef, inputTypeTemplate, dataContext, autoConvert;
 
-  Tracker.nonreactive(function() { 
+  Tracker.nonreactive(function() {
     //don't rerun when data context of element changes, can cause infinite loops
-    
+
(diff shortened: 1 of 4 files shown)


abecks:
I've been using this in production for a week now, with no issues whatsoever.

aldeed:
@abecks thanks. This looks OK to merge, but can you submit a PR to devel branch instead?

KristerV:
@abecks this shit is cool, please resubmit! :)

abecks:
Hey @aldeed I'm trying to work this into the devel branch. My project does not work with devel, I'm getting:

```
Exception in template helper: TypeError: Cannot read property 'objectKeys' of undefined
    at Object.autoFormFieldNames 
```

Form's don't render. I've updated my SimpleSchema to 1.3.3. Any ideas what would cause this?

abecks:
Resubmitted as #1132

PR #1035, opened 2015-06-13
Use the field schema to pass data to a field's context (not as an attribute)

abecks:
This PR allows us to pass data through to a field's context via the field's schema definition. Currently, AutoForm turns all of the schema properties into attributes (omitting a handful i.e. `options`). This is useful when creating schemas on the fly.

```js
summary: {
  type: String,
  autoform: {
    data: {
      someArray: ['apple', 'orange', 'banana'],
      someObj: {
        complex: {
          data: 'structure'
        }
      }
    }
  }
}
```


aldeed:
Since this is specific to input types, it should be within the `afFieldInput` object in the schema. Otherwise it isn't clear which component you're providing the data for.

```js
summary: {
  type: String,
  autoform: {
    afFieldInput: {
      data: {
        someArray: ['apple', 'orange', 'banana'],
        someObj: {
          complex: {
            data: 'structure'
          }
        }
      }
    }
  }
}
```

Also, maybe the data should be added directly to the input context (with _.extend) rather than on the `data` property of the context?

aldeed:
Also, please replace this with a devel branch PR.

abecks:
Recreated as #1133

Issue #827, opened 2015-03-26
Error with nested schema and update-pushArray


ColadaFF:
Hi!
I've been working with the last update of autoForm and and I have the following on my schema: 
```javascript
   "example.$.subdoc": {
        type: Array,
        optional: true
    },
    "example.$.subdoc.$": {
        type: Object,
        optional: true
    },
    "example.$.subdoc.$.SuName": {
        type: String,
    },
```
When I try to create the form with autoform, only show the form if in the scope option I set: ```scope="example.$.subdoc"```, but when I try to update, meteor throws the following exception: ```error: 409, reason: "MinimongoError: The positional operator did not find the match needed from the query", details: undefined```
I also tried with index scope options, like: ```scope="example.0.subdoc```, but in this case the autoForm thows: ```Exception in template helper: Error: Invalid field name: SuName```

Using ```aldeed:autoform@5.1.0```

aldeed:
`scope="example.0.subdoc"` should be correct. It would not work with a `$`. Meteor doesn't support that on the client.

Are you using a `quickForm` or `autoForm`? Can you post the form? I'm not sure why it says "invalid field name".

SenerDemiral:
```
Schemas = {};

Schemas.SS = new SimpleSchema({
  ad: {
    type: String
  },
  aaa: {
    type: [Object],
    optional: true,
    blackbox: true
  },
  "aaa.$.hTakimId": {
    type: String,
    optional: true,
  },
  "aaa.$.vTakimId": {
    type: String,
    optional: true,
  },
  "aaa.$.maclar": {
    type: [Object],
    optional: true,
    blackbox: true
  },
  "aaa.$.maclar.$.hOyuncu": {
    type: String,
    optional: true,
  },
  "aaa.$.maclar.$.vOyuncu": {
    type: String,
    optional: true,
  }
});

SS = new Mongo.Collection('ss');
SS.attachSchema(Schemas.SS);
--------------------------------------
<template name="insSS">
  {{#autoForm collection="SS" id="insSS" type="insert"}}
    <fieldset>
      <legend>SS</legend>
      {{> afQuickField name='ad'}}
    </fieldset>
    <button type="submit" id="btnInsSS" class="btn btn-primary">Insert</button>
  {{/autoForm}}
</template>

<template name="insSS2">
  {{#autoForm collection="SS" id="insSS2" type="insert"}}
    <fieldset>
      <legend>SS2</legend>
[...]

PR #1046, opened 2015-06-17
Fix Error with nested schema and update-pushArray #827

amllano:
The commit fix the issue #827: 

When you have defined an 'update-pushArray' scope using dot notation:

```
{{ #autoForm scope="items.0.item" }}
```

The adjustSchema method behaves improperly because is unable to match the searchString defined as:
`searchString = scope.+'.$.';`

With SimpleSchema keys definition with the form of "items.$.item".

```
_.each(ss.schema(), function (val, key) {
      if (key.indexOf(searchString) === 0) {
        newSchemaDef[key.slice(searchString.length)] = val;
      }
    });
```

The fix: Replaces indexes on scope by '$' during the adjustSchema:
`    var searchString = scope.replace(/[0-9]/g, "$") +'.$.';`

### DIFF
--- a/formTypes/update-pushArray.js
+++ b/formTypes/update-pushArray.js
@@ -40,7 +40,7 @@ AutoForm.addFormType('update-pushArray', {
   },
   adjustSchema: function (ss) {
     var scope = this.form.scope, newSchemaDef = {};
-    var searchString = scope+'.$.';
+    var searchString = scope.replace(/[0-9]/g, "$") +'.$.';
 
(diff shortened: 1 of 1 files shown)


aldeed:
Thanks for the PR. Fixed here: https://github.com/aldeed/meteor-autoform/commit/8544c659ea5235bbbc9f737e170d405c11e23cac

(FYI for the future, please submit PRs to devel branch.)

Issue #859, opened 2015-04-04
formToModifier not being called on a update form


ripdog:
I have a form which I repurpose for both insert and update of appointment items. I have seperate "time" and "date" items in the schema for appointments, the latter of which is hidden from the form and set with the time, which is a string. (The day is set outside of the form).

This logic was contained in docToForm, so 2.1 broke my form updating. No problem, move the logic into a function and call that from both `docToForm` and `docToModifier`, right? I tried that, and now my update field calls neither `docToForm` or `docToModifier`.

Schema: 

```javascript
	date: {
		type: Date,
		label: "Date",
		index: true,
		custom: function() {
			return checkDate(this, true);
		}
	},
	time: {
		type: String,
		label: "Time",
 		regEx: /^[0-2]?\d:\d\d [APap]m|M$/
	},
```

Form: 
```
<template name="insertAppointmentForm">
	{{#autoForm collection=appointmentList doc=currentDoc id="insertAppointmentFormInner" type=currentType}}
...
		{{> afQuickField name='time' value=timePreset type='timePicker'}}
...
		{{> afFieldInput name='date' value=sessionDate type="hidden"}}
	{{/autoForm}}
</template>
```
I have verified that the form is correctly marked as update via `AutoForm.getCurrentDataForForm("insertAppointmentFormInner")`. Validation is default `submitThenKeyup`.

Form JS: 
```javascript
AutoForm.hooks({
	insertAppointmentFormInner: {
...
		formToDoc: function(doc) {
			prepareDocForDB(doc)
		},
		formToModifier: function(doc) {
			console.log(doc);
			prepareDocForDB(doc)
		},
...
   }
});

prepareDocForDB = function(doc) {

	if (typeof doc.time === "string") {
		var datestring = moment(Session.get("date")).format("YYYY-MM-DD ") + doc.time;
		doc.date = moment(datestring, "YYYY-MM-DD hh:mm A").utc().toDate();
	}
	return doc;
}
```

I have set breakpoints on both of the `prepareDocForDB` calls on both Firefox and Chrome, never got called during an update operation.

Thanks so much for your time and hard work! :)

fpoirier1:
I am facing the same issue using the `method-update` form type.

gwendall:
+1

pajooh:
+1

yim1990:
+1

```
Hooks = {
  form: {},
  global: {
    before: {},
    after: {},
    formToDoc: [],
[...]

PR #1056, opened 2015-06-21
Fixed bug preventing use of formToModifier.

aramk:
Added method for creating default hooks.
Added method for getting all hook names.

### DIFF
--- a/autoform-hooks.js
+++ b/autoform-hooks.js
@@ -4,10 +4,7 @@ Hooks = {
   form: {}
 };
 
-/**
- * The names of all supported hooks, excluding "before" and "after".
- * @type {String}
- */
+// The names of all supported hooks, excluding "before" and "after".
 var hookNames = ['formToDoc', 'formToModifier', 'docToForm', 'onSubmit', 'onSuccess', 'onError',
     'beginSubmit', 'endSubmit'];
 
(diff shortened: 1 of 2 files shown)


aramk:
This fixes the issue reported in #859.

autoform-hooks.js
```diff
@@ -1,21 +1,30 @@
...
+  form: {}
+};
+
+/**
+ * @return {String} The names of all supported hooks, excluding "before" and "after".
+ */
+function getHookNames() {
+  return ['formToDoc', 'formToModifier', 'docToForm', 'onSubmit', 'onSuccess',
+      'onError', 'beginSubmit', 'endSubmit'];
+};
```
aldeed: This doesn't need to be a function. You can do `var hookNames = []` and `_.each(hookNames, ...)`.

aldeed:
Thanks @aramk. See my one code comment, and PRs should be submitted to devel branch.

autoform-hooks.js
aramk: Ok, it was mainly to avoid ever mutating the array but since there's no need to override it and we have tests an array should be fine.

aramk:
Thanks @aldeed, I'll re-submit it to devel, seems I can't change the destination once it's created :unamused:

PR #1127, opened 2015-08-10
bug fix - drop down 'select' change not being submitted occasionally in each loop

stinghz:
In each loop,  if the old value of a 'select' change equals to the new value of the previous 'select'  change (on another 'select' control), then this select change won't fire submit event.


aldeed:
This can have side effects, but I'm planning a proper fix for change detection in 6.0.

PR #1133, opened 2015-08-13
Use the field schema to pass data to a field's context (not as an attribute)

abecks:
Recreation of #1035 for devel branch.

This PR allows us to pass data through to a field's context via the field's schema definition. Currently, AutoForm turns all of the schema properties into attributes (omitting a handful i.e. options). This is useful when creating schemas on the fly.

I followed your suggestion and updated it to use the component type and extend the context of the field.

Example usage:
```js
summary: {
  type: String,
  autoform: {
    afFieldInput: {
      data: {
        someArray: ['apple', 'orange', 'banana'],
        someObj: {
          complex: {
            data: 'structure'
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1167,6 +1167,27 @@ summary: {
 
 Tip: Any attribute can instead be provided as a function that returns the attribute's value.
 
+You can pass data structures using the `data` property. They will not be used as attributes, instead they will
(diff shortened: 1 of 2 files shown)


aldeed:
Thanks!

Issue #1163, opened 2015-08-27
afFieldValueIs not working consistently (broken by 5.4.1 release!)


wesyah234:
I have something like this:

```
        {{> afQuickField  name='medicalHistory.field1'}}
        {{#if afFieldValueIs name="medicalHistory.field1" value="Yes"}}
        {{> afQuickField  name='medicalHistory.field2'}}
        {{> afQuickField  name='medicalHistory.field3'}}
        {{/if}}
```

Which should show field2 and field 3 only if the value of field 1 is Yes, however, intermittently, when I open the form, if the value of field1 is Yes, it doesn't show the field2 and field 3.  If I close and reopen the form again it sometimes works properly.

the field1 is a radio button with values of Yes and No, so I can toggle between Yes and No and I see field2 and 3 appearing and disappearing properly, it's just the issue when the value is set to Yes, and sometimes the other fields are hidden, and sometimes they are not.

How should I attempt to troubleshoot this issue?

EDIT: this may be part of the issue, but the fields in question are on a form that is on a tabbed type interface using bootstrap, and the tab is not active when first coming to the page:

```
 <ul class="nav nav-tabs" role="tablist">
```

etc...

wesyah234:
Please see this repo:
https://github.com/wesyah234/quickFormIssue
it will illustrate the problem.
also I posted on the forums: 
https://forums.meteor.com/t/problem-or-bug-with-autoforms-affieldvalueis-helper/8977
thank you for looking at this.

tothand:
Maybe it is related to this issue https://github.com/aldeed/meteor-autoform/issues/861

I know it's an old one but seems to be very similar.

wesyah234:
Yes. I've looked at some of those other issues, and agree.  I also opened another issue: https://github.com/aldeed/meteor-autoform/issues/1051 which was subsequently closed... as duplicate/fixed   At the time I logged #1051  i was reporting that I could get the correct behavior by switching to regular select box instead of radio button, however at this time, with my current issue mentioned above, switching to regular select box does not work.

wesyah234:
just checked my current list of packages for the demo 
```
aldeed:autoform           5.4.1  Easily create forms with automatic insert and update, and automatic reactive validation.
aldeed:collection2        2.4.0  Automatic validation of insert and update operations on the client and server.
iron:router               1.0.9  Routing specifically designed for Meteor
[...]

Issue #1136, opened 2015-08-17
AutoForm.getFieldValue don't return value after upgrade into 5.4.1


chompomonim:
I just upgraded autoform from 5.4.0 into 5.4.1 and now AutoForm.getFieldValue('field_name') returns undefined. After downgrade back into 5.4.0 everything works well.

maxnowack:
same problem here...

rclai:
Same problem here. It seems like putting the form Id into the last parameter helps for some reason.

Every instance where I use the global helper `afFieldValue` also broke.

dard12:
Same behavior here. Breaking on version 5.4.1 but supplying formId works.

For context I'm calling ```AutoForm.getFieldValue``` via a template helper within the ```#autoForm``` block.

abuddenb:
I can confirm that `afFieldValueIs` is broken in 5.4.1.

serkandurusoy:
Yep, I do confirm, too.

I'm declaring autoform options within the schema. Since the schema is not aware of which form (insert/update) it is being called from, how do I obtain a valid formId to pass to getFieldValue so that I can work around the problem?

rodcope1:
I also confirm the issue with 5.4.1.  Unsurprisingly, it's still broken in 5.5.0.

serkandurusoy:
A rather not-so-elegant workaround is:

```javascript
var formId = Autoform.getFormId();
var fieldValue = Autoform.getFieldValue('field', formId);
```

felberj:
For me it looks like it has a problem with getting the value on the first time its rendered (so 'afFieldValueIs/AutoForm.getFieldValue' returns undefined while the form is not fully rendered)

I removed two `!template.view._domrange` checks but now I am stuck as he is trying to read in the DOM what the current value of the field is.

I will look more into it, but unfortunately I don't know Blaze/Meteor very well yet, so if anyone knows how automatically rerun the function when the DOM is ready please answer ;-)

For now I will look more into Autoform and try to see if I somehow get to the form data.

aldeed:
I've merged some PRs. Please see whether this is fixed for you with 5.6.0 release

chompomonim:
Unfortunately it's still broken.

chompomonim:
In 5.7.1 it's still broken. I have to use @serkandurusoy proposed workaround.

maxnowack:
I can confirm that it is still broken.
I use the same workaround and my code looks messy :(

[...]

PR #1195, opened 2015-09-14
Fix 'AutoForm.getFieldValue' when DOM is not ready

felberj:
I am not sure if this is the right way to fix this, but it works for me. #1136 #1163


tcastelli:
any chance of getting this fix for 1.2? (reverting to 5.4.0 now is not possible)

felberj:
Would be cool! @aldeed any plans in accepting this PR?

aldeed:
Thanks. Fixed in a similar but slightly different way in https://github.com/aldeed/meteor-autoform/commit/98ec5f97e449e08a6ebd22097bbbd69cad0f3d75

Issue #1198, opened 2015-09-15
Bug: autosave not working with radio buttons in each loop


challett:
I have set up an each loop to display multiple forms on one page similar to the example found here: http://autoform.meteor.com/update-each except that my forms use select-radio buttons with the same options in each row.

My issue is that when I select the same option in multiple rows of the table, only the first one that I click gets saved.

I have debugged the issue and found that it is caused by an early return here: https://github.com/aldeed/meteor-autoform/blob/396bbcd198000a1dbefc145eabe5b3ff40459070/autoform-events.js#L409

aldeed:
There are several known issues involving that bit of code. This should be fixed eventually by a new system for tracking value changes in 6.0

abernix:
I just submitted will solve this in the most basic way while keeping the existing logic (from #647) that broke it.  It makes the assumption that formId is always set to a string or number, but I believe that's a safe assumption.

I ran into this same problem with a similar form as the example as well – multiple forms with fields with the same name and frequently the same values getting set into them (in my case, quantity fields).  Everything works fine as long as you are entering different values in each field, but with a small character set (0-9) this problem occurs quite frequently.

PR #1199, opened 2015-09-15
Modified check on lastKeyVal to allow for multiple radio buttons with…

challett:
… the same value across multiple forms on the same page

#1198

This changed solved my bug on my local machine.  Not sure if the same problem exists with any other types.


aldeed:
This isn't really a proper fix that handles all use cases, but this is definitely something that needs fixing. I will be completely rewriting the changed value stuff for 6.0 soon.

PR #1237, opened 2015-10-05
Force Array Item containing div to 100% width

dmarkrollins:
In chrome anyways, when you have only one item this div does not always span the full width of the containing panel. So you get this truncated first item but once you add a second item they all then display full width for some reason. Adding this inline style seems to help this template stay true to what I think is the intended format on that first item - 100% width.

Or maybe I am all wrong on this... LOL


aldeed:
Thanks. Fixed in https://github.com/aldeed/meteor-autoform/commit/4649055338f76c552487b4766fc3ee44eeb15afe

PR #1296, opened 2015-11-15
omit in SimpleSchema can now be a function

macrozone:
In SimpleSchema autoform.omit can now be a function. It is a reactive context.

The name of the field is passed as argument. This allows to find sibling or parent fields in nested objects or arrays.

```
new SimpleSchema({
  title: {
    type: String,
    autoform: {
      omit: function(fieldName){
        return true;         
      }
    }
  }
});
```
- [ ] find other / better arguments to pass
- [ ] fix wrong fieldName if field is in array and an element was removed

### DIFF
--- a/autoform-helpers.js
+++ b/autoform-helpers.js
@@ -231,6 +231,10 @@ Template.registerHelper("afFieldNames", function autoFormFieldNames(options) {
       return false;
     }
 
+    if(fieldDefs.autoform && _.isFunction(fieldDefs.autoform.omit) && fieldDefs.autoform.omit(field) === true) {
+      return false;
+    }
+
     // Don't include fields with denyInsert=true when it's an insert form
     if (fieldDefs.denyInsert && form.type === "insert") {
       return false;


macrozone:
I realized that the passed fieldName is wrong, when its a nested field inside an array and if you remove an element from the array.

E.g. you add 2 elements and remove the first. The fieldName, that is passed here is still adresses.1.street instead of adresses.0.street. This is because autoform keeps the removed fieldnames internaly and marks them with "removed: true"

If you want to look up the value of this field (or in my case a silbing field adresses.1.type) with 
AutoForm.getFieldValue("adresses.1.street"), you will receive null or a wrong value

The reason is that AutoForm.getFieldValue will look up the value by creating a MongoObject of the current AutoForm.getFormValues()-document and uss the wrong fieldname ("adresses.1.street") for this lookup.

This document from  AutoForm.getFormValues() will not contain the removed elements in the array, so looking up ""adresses.1.street" will fails as there is only the first element ("adresses.0.street").

A solution might be to actually remove the removed fields from autoforms internal field-store and rename the other fields that belong to the array.

It might be easier to keep the fieldnames with the wrong index and instead make AutoForm.getFieldValue aware of removed fields in arrays.

Related to https://github.com/aldeed/meteor-autoform/issues/914 and https://github.com/aldeed/meteor-autoform/issues/1004

macrozone:
another problem is that when you depend on a field-value in omit(), the whole afFieldNames-helper will re-run, which may have a performance-impact on large forms.

AlmogRnD:
Is there a reason this was never merged?

rafinskipg:
BUMP

Issue #745, opened 2015-03-08
regular expression support on addHooks


aldeed:
See discussion in #298

floriandotorg:
+ 10

ebradbury:
@aldeed I added support for regular expression over here: #1397

Would you mind merging?

PR #1397, opened 2016-02-27
Adds regex support to addHooks

ebradbury:
See issue #745.

This method of supporting regex in `AutoForm.addHooks` relies on adding `onCreated` template hooks to `quickForm` and `autoForm` templates. These hooks are required because the actual ids of the form templates are not known until this point, and therefor the form hook cannot be added util then. These template hooks may have other implications that I'm not aware of so please review and suggest changes.


mchaver:
I'd like to see this merged as well.

Issue #1412, opened 2016-03-07
can't apply custom css class on the afArrayField


maher-samir:
autoform@5.4.0

this can't be working 

```
    param : {
        type : [String],
        autoform: {
            'panelClass': 'customPanelClass',
            'headingClass': 'customHeadingClass',
        }
    }
```

this working fine 

```
    param : {
        type : Object,
        autoform: {
            'panelClass': 'customPanelClass',
            'headingClass': 'customHeadingClass',
        }
    }
```

check here for Object fields 

[templates/bootstrap3/components/afObjectField/afObjectField.html](https://github.com/aldeed/meteor-autoform/blob/2a95d819adb35e14f53a1a048446f19e4e9a072b/templates/bootstrap3/components/afObjectField/afObjectField.html)

```
<template name="afArrayField_bootstrap3">
  <div class="panel panel-default">
    <div class="panel-heading">{{afFieldLabelText name=this.atts.name}}</div>
    {{#if afFieldIsInvalid name=this.atts.name}}
    <div class="panel-body has-error">
      <span class="help-block">{{{afFieldMessage name=this.atts.name}}}</span>
    </div>
    {{/if}}
    <ul class="list-group">
      {{#afEachArrayItem name=this.atts.name minCount=this.atts.minCount maxCount=this.atts.maxCount}}
      <li class="list-group-item autoform-array-item">
        <div>
          <div class="autoform-remove-item-wrap">
            {{#if afArrayFieldHasMoreThanMinimum name=../atts.name minCount=../atts.minCount maxCount=../atts.maxCount}}
            <button type="button" class="btn btn-primary autoform-remove-item"><span class="glyphicon glyphicon-minus"></span></button>
            {{/if}}
          </div>
          <div class="autoform-array-item-body">
            {{> afQuickField name=this.name label=false options=afOptionsFromSchema}}
          </div>
        </div>
      </li>
      {{/afEachArrayItem}}
      {{#if afArrayFieldHasLessThanMaximum name=this.atts.name minCount=this.atts.minCount maxCount=this.atts.maxCount}}
      <li class="list-group-item">
        <button type="button" class="btn btn-primary autoform-add-item" data-autoform-field="{{this.atts.name}}" data-autoform-minCount="{{this.atts.minCount}}" data-autoform-maxCount="{{this.atts.maxCount}}"><span class="glyphicon glyphicon-plus"></span></button>
      </li>
      {{/if}}
    </ul>
  </div>
[...]

PR #1414, opened 2016-03-07
fix issue #1412

maher-samir:
#1412


aldeed:
Merged manually, thanks

PR #1470, opened 2016-05-25
Update README.md

dandalf:
The onSubmit documentation states that only one argument is optionally passed to this.done().  But the onSubmit code example shows that you must use this.done(null, "foo") to pass "foo" to result for onSuccess hook.  I updated the documentation to reflect that example.


aldeed:
Thanks!

Issue #1338, opened 2016-01-06
How to set custom error message with addStickyValidationError()?


renehernandez:
When the form process the error it shows the following sticky message: 
"Unknown validation error"

Even if I put some value as third parameter in the addStickyValidationError call it doesn't change anything.

Any idea how to deal with this?

Thanks in advance

onError hook code:

``` javascript
onError: function(formType, err) {
            var self = this;

            if(err){
                if(err.reason){
                    if (err.reason.indexOf('User') > -1){
                        self.addStickyValidationError('email', 'notFound');
                    }
                    else if (err.reason.indexOf('password') > -1){
                        self.addStickyValidationError('password', 'invalid');
                    }
                }

                AutoForm.validateForm(self.formId);
            }
        }
```

jclay:
I just ran into this issue as well and managed to track down what is going on.

Let's say you have a snippet like this that you're using to notify that user that they're trying to create an account with a duplicate email:
```javascript
AutoForm.hooks({
  insertUser: {
    onError: function (formType, error) {
      if (error.errorType && error.errorType === 'Meteor.Error') {
        console.log(error)
        if (error.reason.match(/exists/)) {
          AutoForm.addStickyValidationError(this.formId, 'email', 'duplicate');
        }
      }
    }
  }
});
```

Our error type is set to `duplicate`. Tracking down the way this is handled reveals that `messageForError` in `SimpleSchema` is eventually called. [Here](https://github.com/aldeed/meteor-simple-schema/blob/11a688d64eb16db570dc534c531500a12e916ac1/simple-schema.js#L960), we can see that it searches for the `errorType` in it's list of known types to try and resolve it to a message. If it cannot find one that matches, this gets called:

```javascript
  if (!message) {
    return "Unknown validation error";
  }
```

The documentation mentions that we can create custom error types [here](https://github.com/aldeed/meteor-simple-schema#customizing-validation-messages).

Now, we see that adding this message to our Schema should solve our problem (and it does!):
```javascript
Schema = {};
Schema.user = new SimpleSchema({
    email: {
        type: String,
        optional: false
    },
    password: {
        type: String,
        optional: false
    }
});

Schema.user.messages({
  "duplicate email": "You're trying to use an email that already exists."
})
```

It's not very clear that this is required from the `auto-form` documentation, so I will look into a fix.

PR #1504, opened 2016-08-15
Clarify the usage of custom messages for stickyErrors

jclay:
This is addressing some confusion as to how to use `addStickyValidationError` with a custom error type and message (see #1338). When attempting to add a sticky error message without first adding a custom error message to your schema instance, you will see "Unknown validation error." This is an attempt to clarify that in the `Readme.`


Pushplaybang:
@jclay  - I'm having trouble with this, constantly get Myschema.messages is not a function, any way around this?

aldeed:
@Pushplaybang This is a bit different in AutoForm 6, which uses a newer NPM SimpleSchema package. It is then `myschema.messageBox` with usage explained here: https://github.com/aldeed/node-message-box

jclay:
Closing this since it seems it's not needed with the latest updates.

PR #1557, opened 2017-01-20
moved dependencies to NPM package simpl-schema

jankapunkt:
I removed the dependency aldeed:simple-schema and included the npm package "simpl-schema" into the app. The tests run green and it's working in my application. 

It does not create a dependency to aldeed:simple-schema in my meteor app (using this package branch and collection2-core) anymore.

Unfortunately I had to include the ecma package, otherwise it would not have been possible to create use the  "simple-schema" on the server without errors (SimpleSchema.extendOptions would then be undefined).

This may cause issues with apps using lower meteor versions, but this is something I don't have the time to figure out.

There was also an issue with the jquery version but I could not find out, wether this was an issue of my app or the package. It may be moved back to the original version if no issues occur.

Unfortunately the pull request is about to be merged to dev, it would be better in my opinion to have an   own branch von deps_move.

Another issue, I could not 100% fix is, that now the app which uses auto form, needs to call

`SimpleSchema.extendOptions(['autoform']);`

at the very beginning, although it is already declared within the package.

### DIFF
--- a/.versions
+++ b/.versions
@@ -1,5 +1,4 @@
 aldeed:autoform@5.8.1
-aldeed:simple-schema@1.1.0
 babel-compiler@5.8.24_1
 babel-runtime@0.1.4
 base64@1.0.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ forms with automatic insert and update events, and automatic reactive validation
 
 ## Change Log
 
+### 5.8.2
+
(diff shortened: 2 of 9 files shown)


jankapunkt:
I found another bug, which I currently fix, so this PR maybe obsolete and maybe replaced (extended) by another PR today or tmr.

jankapunkt:
Close, since the owner will release an official support soon: https://github.com/aldeed/node-simple-schema/issues/44#issuecomment-274831640

Issue #1599, opened 2017-05-30
Unable to reassign the same value after the form data is changed externally.


atf-cguerin:
Hi,

I have just updated to version 6.2.0 because I had the issue with forms not saving described [here](https://github.com/aldeed/meteor-autoform/pull/1381/files) with a set of update forms generated in an `{{#each}}` loop, but there is still one case for me where the `lastKeyVals` trick prevents a client from sending the updated value to the server when the value is changed, as described below.

With two clients A & B (using a value edited in a dropdown ie. 'select' component):
- change the value on A to 'x'
- change the value on B to 'y', which is propagated to A and causes the form to be regenerated
- try changing the value on A to 'x' again: this will be ignored because value stored in lastKeyVals was not updated when the form was regenerated.

**I think I have a fix [attached here](https://github.com/aldeed/meteor-autoform/files/1039138/0001-Fixed-being-unable-to-reassign-the-last-value-change.patch.txt), but could you please review it and merge it if it is correct?**

NB: I also modified the code in `autoFormResetHandler` because it seemed logical given the similar code handling `arrayTracker.resetForm` reset, but if I didn't encounter this version of the bug myself.

PR #1602, opened 2017-06-08
Reset the cache of the last modified key when the form data is reset or updated externally

atf-cguerin:
This is the same patch proposed as a fix for https://github.com/aldeed/meteor-autoform/issues/1599 converted into a pull request after a little more internal testing, in case it can simplify the merging process if it's approved.


Pagebakers:
When can we expect this to be merged?

This has been an issues for a couple of years already where change events of inputs (select/radio) aren't triggered after updating the form sourceDoc.

Pagebakers:
We could also bind the keyVal to the docId if it's available

```javascript
keyVal = key + '___' + keyVal;

if (this.docId) {
keyVal = this.docId + '___' + keyVal
}
if (formId in lastKeyVals && keyVal === lastKeyVals[formId]) {
      return;
    }
```

velatinicolas:
Hello @jankapunkt, I and my work team are currently facing the same issue as described above, and we eventually stumbled upon this pull request which seems to perfectly fix the bug. Do you have any explanation on why it has not been merged? Would it be conceivable to merge it now, even if it has been closed 2 years ago?

jankapunkt:
@velatinicolas I think it has been closed due to being stale for a long time and at that time I was not able to review all PRs. Feel free to reopen and assign me for review, today I will have some time for OS dev 😀

jankapunkt:
@velatinicolas okay I now know why I closed, mainly because it introduces a global-scoped variable which is nowadays considered an anti-pattern.

I could extend the PR by rewriting this to be within the `AutoForm` namespace:

```js
AutoForm._lastKeyVals = {}
```

wdyt?

velatinicolas:
First of all, thanks for your quick answers and propositions! :rocket:

I agree, the global variable bothered me a bit as well. Including it in the `AutoForm` namespace looks great.

But actually, after some tests I made today, for our needs we'd only need to do the fix in the `autoform-events.js` file, because our problem occurs on a form reset. I've not been able to reproduce the bug initially described in the issue https://github.com/Meteor-Community-Packages/meteor-autoform/issues/1599.

As you are the maintainer of this project, I'll let you decide on this. Either we apply the fix in both `autoform-events.js` and `autoForm.js` file, therefore requiring the rework with the namespace ; or we keep the `lastKeyVals` variable declared with `const` and only fix `autoform-events.js`.

velatinicolas:
Thanks for the merge @jankapunkt :rocket:

PR #1693, opened 2020-03-30
V7 merge

### LINKED ISSUE #1692: Upgrade to 7.0.0
We should focus now on updating AutoForm to v7, which has already been implemented by @pouya-eghbali for personal usage.

@pouya-eghbali maybe you just start a PR using your latest stable branch, in the meantime we can all discuss some requirements for review and then start reviewing the PR?

pouya-eghbali:
Merge pouya-eghbali/meteor-autoform


jankapunkt:
So I cloned this and built it into my staging environment after locally were no issues. Our staging users reported no issues with it yet, so it seems to be a good fit :+1: 

The only issues so far are from my end:

* we need to update the documentation
* we need to show how to add Bootstrap 4 templates, mainly because the `imajus:autoform-bootstrap4` is relying on some Template helpers from the Bootstrap3 templates, and I had to create an own package that fixes the (`jkuester:autoform-bootstrap4`)
* we should at least add some unit tests for functions that don't require templating, puppeteer etc. for now (later we may extend this)
* I have tested some of my extensions and they all worked, however the most popular extensions need to be tested, too
* The extensions list may be curated in order to remove abandoned packages or get them into the community if they are very popular

I would offer to review the documentation but the @Meteor-Community-Packages/autoforms team may clone this and try it out locally, too.

pouya-eghbali:
@jankapunkt sorry I'm really really busy these days with everything going remote, I don't know when I can help, maybe in two weeks, but do tell me if you find any issues. Personally I didn't have any issues on my dev machine, staging and on production so far and I don't have any updates for this PR for now.

autoform-api.js
```diff
@@ -82,10 +83,11 @@ AutoForm._globalHooks = Hooks.global;
  * See https://github.com/meteor/meteor/issues/2431
  */
 AutoForm._forceResetFormValues = function autoFormForceResetFormValues(formId) {
-  AutoForm._destroyForm[formId] = AutoForm._destroyForm[formId] || new ReactiveVar(false);
+  AutoForm._destroyForm[formId] =
+    AutoForm._destroyForm[formId] || new ReactiveVar(false);
```
jankapunkt, line 87: I wonder if this formatting is easier to read than the line above.

autoform-api.js
```diff
@@ -132,12 +134,25 @@ AutoForm.getDefaultTemplate = function autoFormGetDefaultTemplate() {
  * @param {String} type
  * @param {String} template
  */
-AutoForm.setDefaultTemplateForType = function autoFormSetDefaultTemplateForType(type, template) {
+AutoForm.setDefaultTemplateForType = function autoFormSetDefaultTemplateForType(
+  type,
+  template
+) {
```
jankapunkt, line 140: Same here. I think this makes sense when there are many parameters, but with two parameters it seems to be even harder to read than the one-liner.

autoform-api.js
```diff
@@ -132,12 +134,25 @@ AutoForm.getDefaultTemplate = function autoFormGetDefaultTemplate() {
...
+      "setDefaultTemplateForType can't set default template to \"" +
+        template +
+        '" for type "' +
+        type +
+        '" because there is no defined template with the name "' +
+        type +
+        "_" +
+        template +
+        '"'
+    );
```
jankapunkt, line 155: Can't we use [template-literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here? I think we should use template-literals for any string concatenation.

autoform-api.js
```diff
@@ -151,7 +166,9 @@ AutoForm.setDefaultTemplateForType = function autoFormSetDefaultTemplateForType(
  *
  * Reactive.
  */
-AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(type) {
+AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
+  type
+) {
```
jankapunkt, line 171: Also too much to break for one char, IMO

autoform-api.js
```diff
@@ -180,61 +197,93 @@ AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
  * 6. Default template, as set by AutoForm.setDefaultTemplate.
  * 7. Built-in default template, currently bootstrap-3.
  */
-AutoForm.getTemplateName = function autoFormGetTemplateName(templateType, templateName, fieldName, skipExistsCheck) {
+AutoForm.getTemplateName = function autoFormGetTemplateName(
+  templateType,
+  templateName,
+  fieldName,
+  skipExistsCheck
+) {
```
jankapunkt, line 205: This for example makes total sense to me to break, this much more readable than in one line

autoform-api.js
```diff
@@ -180,61 +197,93 @@ AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
...
-    return templateType + '_' + schemaAutoFormDefs[templateType].template;
+  if (
+    schemaAutoFormDefs &&
+    schemaAutoFormDefs[templateType] &&
+    schemaAutoFormDefs[templateType].template &&
+    templateExists(
+      templateType + "_" + schemaAutoFormDefs[templateType].template
+    )
+  ) {
+    return templateType + "_" + schemaAutoFormDefs[templateType].template;
```
jankapunkt, line 242: Can we use template-literals here?

autoform-api.js
```diff
@@ -180,61 +197,93 @@ AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
...
 
   // Fallback #2: autoform.template from the schema
-  if (schemaAutoFormDefs && schemaAutoFormDefs.template && templateExists(templateType + '_' + schemaAutoFormDefs.template)) {
-    return templateType + '_' + schemaAutoFormDefs.template;
+  if (
+    schemaAutoFormDefs &&
+    schemaAutoFormDefs.template &&
+    templateExists(templateType + "_" + schemaAutoFormDefs.template)
+  ) {
+    return templateType + "_" + schemaAutoFormDefs.template;
```
jankapunkt, line 251: Can we use template-literals here?

autoform-api.js
```diff
@@ -180,61 +197,93 @@ AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
...
   // Fallback #3: template-<componentType> attribute on an ancestor component within the same form
-  let templateFromAncestor = AutoForm.findAttribute('template-' + templateType);
-  if (templateFromAncestor && templateExists(templateType + '_' + templateFromAncestor)) {
-    return templateType + '_' + templateFromAncestor;
+  let templateFromAncestor = AutoForm.findAttribute("template-" + templateType);
+  if (
+    templateFromAncestor &&
+    templateExists(templateType + "_" + templateFromAncestor)
+  ) {
+    return templateType + "_" + templateFromAncestor;
```
jankapunkt, line 260: Can we use template-literals here?

autoform-api.js
```diff
@@ -180,61 +197,93 @@ AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
...
   // Fallback #4: template attribute on an ancestor component within the same form
-  templateFromAncestor = AutoForm.findAttribute('template');
-  if (templateFromAncestor && templateExists(templateType + '_' + templateFromAncestor)) {
-    return templateType + '_' + templateFromAncestor;
+  templateFromAncestor = AutoForm.findAttribute("template");
+  if (
+    templateFromAncestor &&
+    templateExists(templateType + "_" + templateFromAncestor)
+  ) {
+    return templateType + "_" + templateFromAncestor;
```
jankapunkt, line 269: Can we use template-literals here?

autoform-api.js
```diff
@@ -180,61 +197,93 @@ AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
...
-  if (defaultTemplateForType && templateExists(templateType + '_' + defaultTemplateForType)) {
-    return templateType + '_' + defaultTemplateForType;
+  const defaultTemplateForType = AutoForm.getDefaultTemplateForType(
+    templateType
+  );
+  if (
+    defaultTemplateForType &&
+    templateExists(templateType + "_" + defaultTemplateForType)
+  ) {
+    return templateType + "_" + defaultTemplateForType;
```
jankapunkt, line 280: Can we use template-literals here?

autoform-api.js
```diff
@@ -180,61 +197,93 @@ AutoForm.getDefaultTemplateForType = function autoFormGetDefaultTemplateForType(
...
   }
 
   // Fallback #6: Default template, as set by AutoForm.setDefaultTemplate
   const defaultTemplate = AutoForm.getDefaultTemplate();
-  if (defaultTemplate && templateExists(templateType + '_' + defaultTemplate)) {
-    return templateType + '_' + defaultTemplate;
+  if (defaultTemplate && templateExists(templateType + "_" + defaultTemplate)) {
+    return templateType + "_" + defaultTemplate;
```
jankapunkt, line 286: Can we use template-literals here?

autoform-api.js
```diff
@@ -343,18 +401,20 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
+    if (clean) {
+      ss.clean(insertDoc, {
+        isModifier: false,
+        getAutoValues: false,
+        filter: filter,
+        autoConvert: autoConvert,
+        trimStrings: trimStrings,
+        mutate: true
+      });
+    }
```
jankapunkt, line 413: Is this new behavior added to the documentation already?

autoform-api.js
```diff
@@ -365,21 +425,23 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
     // Do not add autoValues at this stage.
     updateDoc = AutoForm.Utility.docToModifier(doc, {
```
jankapunkt, line 426: I question to this - does it make sense to add a flag, that prevents returning one of `insertDoc` or `updateDoc`? I think it is not necessary to generate both all the time, since each form serves one purpose. To be backward compatible this could be done by checking explicit `false` and otherwise by default return both, as it already does. wdyt?

autoform-api.js
```diff
@@ -398,19 +460,70 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
 
+/**
+ * @method AutoForm.resetValueCache
+ * @public
+ * @param {String} [formId] The `id` attribute of the `autoForm` you want current values for. Default is the closest form from the current context.
+ * @param {String} [fieldName] The field `name` you want to reset the cache of.
+ * @return {Any|undefined}
+ *
+ * Reset the cache and mark all fields as changed
+ */
```
jankapunkt, line 471: Is this updated in the API.md already?

autoform-api.js
```diff
@@ -398,19 +460,70 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
+  });
+
+  template.formValues = template.formValues || {};
+
+  if (fieldName) {
+    if (template.formValues[fieldName]) {
+      template.formValues[fieldName].isMarkedChanged = true;
+    }
+    // reset ancestors
+    const { ancestors } = fieldName
```
jankapunkt, line 485: Maybe adding a comment here about the structure of `ancestors`, that is about to be processed?

autoform-api.js
```diff
@@ -398,19 +460,70 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
+      .split(".")
+      .slice(0, -1)
+      .reduce(
+        ({ ancestors, parent }, sub) => {
+          parent = parent ? parent + "." + sub : sub;
```
jankapunkt, line 490: Can we use template-literals here?

autoform-api.js
```diff
@@ -398,19 +460,70 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
+
+  if (fieldName) {
+    if (template.formValues[fieldName]) {
+      template.formValues[fieldName].isMarkedChanged = true;
+    }
+    // reset ancestors
+    const { ancestors } = fieldName
+      .split(".")
+      .slice(0, -1)
+      .reduce(
```
jankapunkt, line 488: Also please a short comment on the reduce function, I think they are often the hardest to understand when their behavior is not described

autoform-api.js
```diff
@@ -398,19 +460,70 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
+        ({ ancestors, parent }, sub) => {
+          parent = parent ? parent + "." + sub : sub;
+          ancestors.push(parent);
+          return { ancestors, parent };
+        },
+        { ancestors: [], parent: "" }
+      );
+
+    for (ancestor of ancestors) {
```
jankapunkt, line 497: Shouldn't we define for loop variables either using `let` or `const`?

```javascript
 for (const ancestor of ancestors) { ... }
```
Is there a reason to not use `ancestors.forEach` here?

autoform-api.js
```diff
@@ -398,19 +460,70 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
+    });
+  }
+};
+
 /**
  * @method AutoForm.getFieldValue
  * @public
  * @param {String} fieldName The name of the field for which you want the current value.
  * @param {String} [formId] The `id` attribute of the `autoForm` you want current values for. Default is the closest form from the current context.
+ * @param {Boolean} [clean] Indicates if the method should clean the values or not. Default is true.
```
jankapunkt, line 514: May it be, that we wll have more options parameters here in the future? If so, it would make sense to make the third parameter an Object to allow non-breaking adds of arbitrary options:

```javascript
 function autoFormGetFieldValue(fieldName, formId, { clean = true } = {}) { ... }
```

autoform-api.js
```diff
@@ -398,19 +460,70 @@ AutoForm.getFormValues = function autoFormGetFormValues(formId, template, ss, ge
...
  *
  * Returns the value of the field (the value that would be used if the form were submitted right now).
  * This is a reactive method that will rerun whenever the current value of the requested field changes. Return value will be undefined if the field is not currently rendered.
  */
-AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
+AutoForm.getFieldValue = function autoFormGetFieldValue(
+  fieldName,
+  formId,
+  clean = true
+) {
```
jankapunkt, line 524: Is the new paramter structure added to API.md?

autoform-api.js
```diff
@@ -423,16 +536,74 @@ AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
...
 
   var mDoc = new MongoObject(doc);
-  return mDoc.getValueForKey(fieldName);
+  var value = mDoc.getValueForKey(fieldName);
+
+  template.formValues[fieldName].cachedValue = value;
+  template.formValues[fieldName].isMarkedChanged = false;
+
+  return value;
+};
```
jankapunkt, line 560: Could you please add some method-internal comments on how `isMarkedChanged` is introducing new behavior to this mathod?

autoform-api.js
```diff
@@ -423,16 +536,74 @@ AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
...
+ * @method AutoForm.setFieldValue
+ * @public
+ * @param {String} fieldName The name of the field for which you want to set the current value.
+ * @param {Any} value Value of the field
+ * @param {String} [formId] The `id` attribute of the `autoForm` you want current values for. Default is the closest form from the current context.
+ * @return {Any|undefined}
+ *
+ * Sets the value for a field, resets the cache and emits changed events.
+ */
+AutoForm.setFieldValue = function autoFormSetFieldValue(
```
jankapunkt, line 572: Is this already added to API.md?

autoform-api.js
```diff
@@ -423,16 +536,74 @@ AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
...
+  fieldName,
+  value,
+  formId
+) {
+  const mDoc =
+    Tracker.nonreactive(function() {
+      return AutoForm.reactiveFormData.sourceDoc(formId);
+    }) || new MongoObject({});
+
+  const keyExists = Boolean(mDoc.getInfoForKey(fieldName));
```
jankapunkt, line 582: Are `mDoc` and  `getInfoForKey` things, that are definitely always present or should we check like

```javascript
const keyExists = Boolean(mDoc && mDoc.getInfoForKey && mDoc.getInfoForKey(fieldName));
```

autoform-api.js
```diff
@@ -423,16 +536,74 @@ AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
...
+/**
+ * @method AutoForm.setFormValues
+ * @public
+ * @param {Any} value Value of the form
+ * @param {String} [formId] The `id` attribute of the `autoForm` you want current values for. Default is the closest form from the current context.
+ * @return {Any|undefined}
+ *
+ * Sets the value for a field, resets the cache and emits changed events.
+ */
+AutoForm.setFormValues = function autoFormSetFormValues(value, formId) {
```
jankapunkt, line 605: Is this added to `API.md`?

autoform-api.js
```diff
@@ -443,22 +614,26 @@ AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
  *
  * Returns the name of the template used to render the element.
  */
-AutoForm.getInputTypeTemplateNameForElement = function autoFormGetInputTypeTemplateNameForElement(element) {
+AutoForm.getInputTypeTemplateNameForElement = function autoFormGetInputTypeTemplateNameForElement(
+  element
```
jankapunkt, line 618: This could remain in one line IMO

autoform-api.js
```diff
@@ -443,22 +614,26 @@ AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
...
   // the template contains a block helper like if, with, each,
   // then look up the view chain until we arrive at a template
-  while (view &&
-         view.name.indexOf('Template.') !== 0 &&
-         view.name.indexOf('BlazeComponent.') !== 0) {
+  while (
+    view &&
+    view.name.indexOf("Template.") !== 0 &&
+    view.name.indexOf("BlazeComponent.") !== 0
+  ) {
```
jankapunkt, line 629: We could also use `view.name.includes(someString)` -> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

autoform-api.js
```diff
@@ -443,22 +614,26 @@ AutoForm.getFieldValue = function autoFormGetFieldValue(fieldName, formId) {
...
     view = view.originalParentView || view.parentView;
   }
 
   if (!view) return;
 
   // View names have "Template." or "BlazeComponent." at the beginning so we slice that off.
-  return view.name.slice(view.name.indexOf('.') + 1);
+  return view.name.slice(view.name.indexOf(".") + 1);
 };
```
jankapunkt, line 637: Maybe use `view.name.includes`?

autoform-api.js
```diff
@@ -496,16 +680,18 @@ AutoForm.getInputValue = function autoFormGetInputValue(element, ss) {
...
-  var lastUnderscore = inputTypeTemplate.lastIndexOf('_');
+  var lastUnderscore = inputTypeTemplate.lastIndexOf("_");
   if (lastUnderscore !== -1) {
     inputTypeTemplate = inputTypeTemplate.slice(0, lastUnderscore);
   }
 
   // Figure out what registered input type was used to render this element
-  typeDef = _.where(AutoForm._inputTypeDefinitions, { template: inputTypeTemplate })[0];
+  typeDef = Object.values(AutoForm._inputTypeDefinitions).filter(
+    def => def.template == inputTypeTemplate
```
jankapunkt, line 690: We should use always compare using  `===` instead of `==` since `==` can produce many false-positives/false-negatives

autoform-api.js
```diff
@@ -591,7 +787,8 @@ AutoForm.validateField = function autoFormValidateField(formId, fieldName, skipE
  */
 AutoForm.validateForm = function autoFormValidateForm(formId) {
   var form = AutoForm.getCurrentDataForForm(formId);
-  var formDoc, formType = form.type;
+  var formDoc,
+    formType = form.type;
```
jankapunkt, line 791: We should use always `const` or `let` since `var` cannot define variables in [block-scope](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let).

Also for readability it may be good to define each variable in a separate line like

```javascript
const form = AutoForm.getCurrentDataForForm(formId);
const formType = form.type;
let formDoc = undefined;

...
```

autoform-api.js
```diff
@@ -654,21 +854,21 @@ AutoForm.findAttribute = function autoFormFindAttribute(attrName) {
...
+      if (attrName !== "template" || !("data" in viewData)) {
         val = viewData[attrName];
       }
     }
   }
 
   // Loop
   view = Blaze.currentView;
-  while (val === undefined && view && view.name !== 'Template.autoForm') {
+  while (val === undefined && view && view.name !== "Template.autoForm") {
```
jankapunkt, line 865: If we often check against the same String, it makes sense to create a short helper

```javascript
const isAutoFormTemplate = view => view && view.name === "Template.autoForm";
```

autoform-api.js
```diff
@@ -683,8 +883,13 @@ AutoForm.findAttribute = function autoFormFindAttribute(attrName) {
...
  */
-AutoForm.findAttributesWithPrefix = function autoFormFindAttributesWithPrefix(prefix) {
-  var result = {}, view, viewData, searchObj;
+AutoForm.findAttributesWithPrefix = function autoFormFindAttributesWithPrefix(
+  prefix
+) {
+  var result = {},
+    view,
+    viewData,
+    searchObj;
```
jankapunkt, line 892: Please let's define variables as either `let` or `const`, also this makes the code more verbose to understand roles of the variables.

autoform-api.js
```diff
@@ -695,9 +900,9 @@ AutoForm.findAttributesWithPrefix = function autoFormFindAttributesWithPrefix(pr
...
     }
-    // We need an isArray check, too because _.isObject([{}]) comes back true
-    if (_.isObject(searchObj) && !_.isArray(searchObj)) {
-      _.each(searchObj, function (v, k) {
+    // We need an isArray check, too because isObject([{}]) comes back true
+    if (isObject(searchObj) && !Array.isArray(searchObj)) {
+      Object.entries(searchObj).forEach(function([k, v]) {
         if (k.indexOf(prefix) === 0) {
           result[k.slice(prefix.length)] = v;
         }
```
jankapunkt, line 908: Alternative suggestion:

```javascript
Object.entries(searchObj).forEach(function([key, value]) {
  if (key.includes(prefix)) {
    const suffix = key.slice(prefix.length)
    result[suffix] = value;
  }
```

autoform-api.js
```diff
@@ -707,13 +912,13 @@ AutoForm.findAttributesWithPrefix = function autoFormFindAttributesWithPrefix(pr
 
   // Loop
   view = Blaze.currentView;
-  while (view && view.name !== 'Template.autoForm') {
+  while (view && view.name !== "Template.autoForm") {
```
jankapunkt, line 915: Could use the helper I added some comments before

autoform-api.js
```diff
@@ -707,13 +912,13 @@ AutoForm.findAttributesWithPrefix = function autoFormFindAttributesWithPrefix(pr
...
     checkView();
     view = view.originalParentView || view.parentView;
   }
 
   // If we've reached the form, check there, too
-  if (view && view.name === 'Template.autoForm') {
+  if (view && view.name === "Template.autoForm") {
```
jankapunkt, line 921: Could use the helper I added some comments before

autoform-api.js
```diff
@@ -730,8 +935,8 @@ AutoForm.findAttributesWithPrefix = function autoFormFindAttributesWithPrefix(pr
 AutoForm.debug = function autoFormDebug() {
   AutoForm._debug = true;
   AutoForm.addHooks(null, {
-    onError: function (operation, error) {
-      console.log('Error in ' + this.formId, operation, error);
```
jankapunkt, line 734: Let's use template literals here

autoform-api.js
```diff
@@ -924,11 +1129,11 @@ AutoForm.viewForForm = function (formId) {
     view = Blaze.getView(formElement);
   } catch (err) {}
 
-  while (view && view.name !== 'Template.autoForm') {
+  while (view && view.name !== "Template.autoForm") {
```
jankapunkt, line 1132: Could use the helper I defined in some comments before

autoform-api.js
```diff
@@ -924,11 +1129,11 @@ AutoForm.viewForForm = function (formId) {
...
     view = view.originalParentView || view.parentView;
   }
 
-  if (!view || view.name !== 'Template.autoForm') {
+  if (!view || view.name !== "Template.autoForm") {
```
jankapunkt, line 1136: Could use the helper I defined in some comments before

autoform-api.js
```diff
@@ -1002,13 +1207,13 @@ AutoForm.getCurrentDataForForm = function (formId) {
  * You can call this without a formId from within a helper and
  * the data for the nearest containing form will be returned.
  */
-AutoForm.getCurrentDataPlusExtrasForForm = function (formId) {
+AutoForm.getCurrentDataPlusExtrasForForm = function(formId) {
   var data = AutoForm.getCurrentDataForForm(formId);
 
-  data = _.clone(data);
+  data = { ...data };
```
jankapunkt, line 1213: Maybe adding a short comment on why using the Object-spread and not `Object.assign` here

autoform-api.js
```diff
@@ -1064,18 +1269,23 @@ AutoForm.getFormId = function () {
...
  */
-AutoForm.selectFirstInvalidField = function selectFirstInvalidField(formId, ss) {
-  var ctx = ss.namedContext(formId), template, fields;
+AutoForm.selectFirstInvalidField = function selectFirstInvalidField(
+  formId,
+  ss
+) {
+  var ctx = ss.namedContext(formId),
+    template,
+    fields;
```
jankapunkt, line 1278: Please let's use `let` and `const` for more verbosity of the code

autoform-api.js
```diff
@@ -1064,18 +1269,23 @@ AutoForm.getFormId = function () {
...
   if (!ctx.isValid()) {
     template = AutoForm.templateInstanceForForm(formId);
     fields = getAllFieldsInForm(template);
-    fields.each(function () {
+    fields.each(function() {
       var f = $(this);
```
jankapunkt, line 1283: can `f` be renamed? One-char variables make the code really hard to read IMO

autoform-api.js
```diff
@@ -1176,9 +1403,9 @@ AutoForm._validateFormDoc = function validateFormDoc(doc, isModifier, formId, ss
 
     // Add sticky errors for all keys if any
     var stickyErrors = AutoForm.templateInstanceForForm(formId)._stickyErrors;
-    if (!_.isEmpty(stickyErrors)) {
+    if (Object.keys(stickyErrors).length) {
       isValid = false;
-      stickyErrors = _.map(stickyErrors, function (obj, k) {
+      stickyErrors = Object.entries(stickyErrors).map(function([k, obj]) {
```
jankapunkt, line 1408: can we make `k` to `key`?

common.js
```diff
@@ -0,0 +1,10 @@
+export const isObject = obj => toString.call(obj) == '[object Object]' || Array.isArray(obj) || isFunction(obj)
+export const isFunction = obj => toString.call(obj) == '[object Function]'
+
+export const throttle = (fn, limit) => {
+  let timeout
+  return (...args) => {
+    clearTimeout(timeout)
+    timeout = setTimeout(() => fn(...args), limit)
+  }
+}
```
jankapunkt, line 10: Please add a new line at the end of the file

common.js
```diff
@@ -0,0 +1,10 @@
+export const isObject = obj => toString.call(obj) == '[object Object]' || Array.isArray(obj) || isFunction(obj)
```
jankapunkt, line 1: Shouldn't `isObject` return false if it's an array or a function?

autoform-arrays.js
```diff
@@ -44,29 +49,47 @@ ArrayTracker.prototype.ensureField = function atEnsureField(formId, field) {
...
+  var arrayCount = Math.max(range.minCount, docCount == null ? 1 : docCount);
 
-// If this is an array of objects, collect names of object props
+  // If this is an array of objects, collect names of object props
   var childKeys = [];
-  if (AutoForm.Utility.getFieldDefinition(ss, field + '.$').type === Object) {
-    childKeys = ss.objectKeys(AutoForm.Utility.makeKeyGeneric(field) + '.$');
-	}
+  if (AutoForm.Utility.getFieldDefinition(ss, field + ".$").type === Object) {
+    childKeys = ss.objectKeys(AutoForm.Utility.makeKeyGeneric(field) + ".$");
```
jankapunkt, line 73: Can we use template literals to concat Strings?

autoform-arrays.js
```diff
@@ -44,29 +49,47 @@ ArrayTracker.prototype.ensureField = function atEnsureField(formId, field) {
...
+      i,
+      childKeys,
+      overrideMinCount,
+      overrideMaxCount
+    );
     loopArray.push(loopCtx);
+    collection.insert(loopCtx);
   }
 
+  self.info[formId][field].collection = collection;
```
jankapunkt, line 92: Can we be sure that there is always the ability to access these arrays without error?

autoform-arrays.js
```diff
@@ -7,7 +7,12 @@ ArrayTracker = function afArrayTracker() {
   self.info = {};
 };
 
-ArrayTracker.prototype.getMinMax = function atGetMinMax(ss, field, overrideMinCount, overrideMaxCount) {
+ArrayTracker.prototype.getMinMax = function atGetMinMax(
```
jankapunkt, line 10: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -44,29 +49,47 @@ ArrayTracker.prototype.ensureField = function atEnsureField(formId, field) {
   if (!self.info[formId][field]) self.resetField(formId, field);
 };
 
-ArrayTracker.prototype.initField = function atInitField(formId, field, ss, docCount, overrideMinCount, overrideMaxCount) {
+ArrayTracker.prototype.initField = function atInitField(
```
jankapunkt, line 52: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -92,13 +119,20 @@ ArrayTracker.prototype.resetField = function atResetField(formId, field) {
 
 ArrayTracker.prototype.resetForm = function atResetForm(formId) {
```
jankapunkt, line 120: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -92,13 +119,20 @@ ArrayTracker.prototype.resetField = function atResetField(formId, field) {
...
   var self = this;
-  _.each(self.info[formId], function (info, field) {
+  Object.keys(self.info[formId] || {}).forEach(function(field) {
     self.resetField(formId, field);
-	});
+  });
 };
 
 ArrayTracker.prototype.untrackForm = function atUntrackForm(formId) {
```
jankapunkt, line 127: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
   return self.info[formId][field].count;
 };
 
-ArrayTracker.prototype.getVisibleCount = function atGetVisibleCount(formId, field) {
+ArrayTracker.prototype.getVisibleCount = function atGetVisibleCount(
```
jankapunkt, line 160: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
+  field
+) {
   var self = this;
   self.ensureField(formId, field);
   self.info[formId][field].deps.depend();
   return self.info[formId][field].visibleCount;
 };
 
-ArrayTracker.prototype.isFirstFieldlVisible = function atIsFirstFieldlVisible(formId, field, currentIndex) {
+ArrayTracker.prototype.isFirstFieldlVisible = function atIsFirstFieldlVisible(
```
jankapunkt, line 170: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
+  ) {
     return !currentField.removed;
-	});
-	return (firstVisibleField && firstVisibleField.index === currentIndex);
+  });
+  return firstVisibleField && firstVisibleField.index === currentIndex;
 };
 
-ArrayTracker.prototype.isLastFieldlVisible = function atIsLastFieldlVisible(formId, field, currentIndex) {
+ArrayTracker.prototype.isLastFieldlVisible = function atIsLastFieldlVisible(
```
jankapunkt, line 186: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
-  return (lastVisibleField && lastVisibleField.index === currentIndex);
+  var lastVisibleField = self.info[formId][field].array
+    .filter(function(currentField) {
+      return !currentField.removed;
+    })
+    .pop();
+  return lastVisibleField && lastVisibleField.index === currentIndex;
 };
 
-ArrayTracker.prototype.addOneToField = function atAddOneToField(formId, field, ss, overrideMinCount, overrideMaxCount) {
```
jankapunkt, line 153: Please add a short jsDoc comment what this method does.

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
     var childKeys = [];
-    if (AutoForm.Utility.getFieldDefinition(ss, field + '.$').type === Object) {
-      childKeys = ss.objectKeys(AutoForm.Utility.makeKeyGeneric(field) + '.$');
-	  }
-
-    var loopCtx = createLoopCtx(formId, field, i, childKeys, overrideMinCount, overrideMaxCount);
-
+    if (AutoForm.Utility.getFieldDefinition(ss, field + ".$").type === Object) {
+      childKeys = ss.objectKeys(AutoForm.Utility.makeKeyGeneric(field) + ".$");
+    }
```
jankapunkt, line 225: Can we use template literals to concat strings?

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
 
   if (!self.info[formId][field].array) return;
 
   var currentCount = self.info[formId][field].visibleCount;
-  var maxCount = self.getMinMax(ss, field, overrideMinCount, overrideMaxCount).maxCount;
+  var maxCount = self.getMinMax(ss, field, overrideMinCount, overrideMaxCount)
+    .maxCount;
 
   if (currentCount < maxCount) {
     var i = self.info[formId][field].array.length;
```
jankapunkt, line 219: please use `let` or `const` because `var`is not block scoped here. Coould we also rename `i` to something more verbose?

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
+  overrideMaxCount
+) {
   var self = this;
   self.ensureField(formId, field);
 
   if (!self.info[formId][field].array) return;
 
   var currentCount = self.info[formId][field].visibleCount;
-  var minCount = self.getMinMax(ss, field, overrideMinCount, overrideMaxCount).minCount;
+  var minCount = self.getMinMax(ss, field, overrideMinCount, overrideMaxCount)
```
jankapunkt, line 260: Please use `let` or `const` for variables

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
+    .minCount;
```
jankapunkt, line 261: Could also be written as

```javascript
const { minCount } = self.getMinMax(...)
```

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
     arrayFieldName: field,
-    name:           field + '.' + index,
-    index:          index,
-    minCount:       overrideMinCount,
-    maxCount:       overrideMaxCount
+    name: field + "." + index,
+    index: index,
+    minCount: overrideMinCount,
+    maxCount: overrideMaxCount
   };
```
jankapunkt, line 295: Can we use template literals to concat strings and avoid `var`?

autoform-arrays.js
```diff
@@ -123,95 +157,148 @@ ArrayTracker.prototype.getCount = function atGetCount(formId, field) {
...
 
   // If this is an array of objects, add child key names under loopCtx.current[childName] = fullKeyName
   if (childKeys.length) {
     loopCtx.current = {};
-    _.each(childKeys, function (k) {
-      loopCtx.current[k] = field + '.' + index + '.' + k;
+    childKeys.forEach(function(k) {
+      loopCtx.current[k] = field + "." + index + "." + k;
     });
```
jankapunkt, line 302: Can we use template literals and write `k` as `key`?

autoform-formdata.js
```diff
@@ -6,7 +6,7 @@
  * keep their type upon retrieval.
  */
 
-FormData = function () {
+FormData = function() {
```
jankapunkt, line 9: does it make sense to rewrite this as an es6 class? 

```javascript
FormData = class FormData { ... }
```

autoform-helpers.js
```diff
@@ -107,7 +109,10 @@ Template.registerHelper('afFieldValueContains', function autoFormFieldValueConta
   options = parseOptions(options, 'afFieldValueContains');
 
   var currentValue = AutoForm.getFieldValue(options.name, options.formId);
-  return _.isArray(currentValue) && (_.contains(currentValue, options.value) || options.values && _.intersection(currentValue, options.values.split(',')));
+  return Array.isArray(currentValue)
+    && (currentValue.includes(options.value)
+      || options.values
+      && options.values.split(',').filter(item => currentValue.includes(item)));
 });
```
jankapunkt, line 116: There should be a round bracket around:

```
...
|| (options.values && options.values.split().filter() )
```

and indentation is 2 spaces too far right

autoform-helpers.js
```diff
@@ -158,22 +163,22 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
...
+      fieldList = fieldList.filter(function filterFieldsByName(field) {
         return field.indexOf(namePlusDot) === 0;
       });
     }
 
     // If top level fields, be sure to remove any with $ in them
     else {
-      fieldList = _.filter(fieldList, function filterArrayFields(field) {
+      fieldList = fieldList.filter(function filterArrayFields(field) {
         return (field.slice(-2) !== '.$' && field.indexOf('.$.') === -1);
```
jankapunkt, line 182: Maybe use `&& !field.includes('.$.')`?

autoform-helpers.js
```diff
@@ -186,17 +191,17 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
     // "city" field within that, but if you instead do `fields="address.city"`
     // we will use a single field for the city, with no afObjectField
     // template around it.
-    fieldList = _.reject(fieldList, function (field) {
+    fieldList = fieldList.filter(function (field) {
       var lastDotPos = field.lastIndexOf('.');
       if (lastDotPos === -1) {
-        return false; // keep
+        return true; // keep
```
jankapunkt, line 197: Is there a related issue that has been fixed due to this?

autoform-helpers.js
```diff
@@ -186,17 +191,17 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
     // "city" field within that, but if you instead do `fields="address.city"`
     // we will use a single field for the city, with no afObjectField
     // template around it.
-    fieldList = _.reject(fieldList, function (field) {
+    fieldList = fieldList.filter(function (field) {
       var lastDotPos = field.lastIndexOf('.');
```
jankapunkt, line 195: Please use `const`

autoform-helpers.js
```diff
@@ -186,17 +191,17 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
...
-        return false; // keep
+        return true; // keep
       }
 
       var parentField = field.slice(0, lastDotPos);
       if (parentField.slice(-2) === '.$') {
         parentField = parentField.slice(0, -2);
       }
-      return _.contains(fieldList, parentField) && parentField !== name && parentField !== genericName;
+      return !fieldList.includes(parentField) || parentField === name || parentField === genericName
```
jankapunkt, line 204: Does this change fix an issue? Please add as comment if so

autoform-helpers.js
```diff
@@ -209,7 +214,7 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
     if (name) {
       // Tack child field name on to end of parent field name. This
       // ensures that we keep the desired array index for array items.
-      fieldList = _.map(fieldList, function (field) {
+      fieldList = fieldList.map(function (field) {
         return name + '.' + field;
```
jankapunkt, line 218: Can we use template-literals here?

autoform-helpers.js
```diff
@@ -240,6 +245,10 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
       return false;
     }
 
+    if (fieldDefs.autoform && isFunction(fieldDefs.autoform.omit) && fieldDefs.autoform.omit(field) === true) {
```
jankapunkt, line 248: Can you please add a comment if this new introduced behavior? Also - does it fix an issue? If so, please comment, too.

autoform-helpers.js
```diff
@@ -219,15 +224,15 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
...
-    fieldList = _.reject(fieldList, function (f) {
-      return _.contains(omitFields, AutoForm.Utility.makeKeyGeneric(f));
+    fieldList = fieldList.filter(function (f) {
+      return !omitFields.includes(AutoForm.Utility.makeKeyGeneric(f));
     });
   }
 
   // Filter out fields we never want
-  fieldList = _.filter(fieldList, function shouldIncludeField(field) {
+  fieldList = fieldList.filter(function shouldIncludeField(field) {
```
jankapunkt, line 235: If `fieldList` is reassigned, it should be declared using `let`

autoform-helpers.js
```diff
@@ -219,15 +224,15 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
...
     var fieldDefs = AutoForm.Utility.getFieldDefinition(ss, field);
```
jankapunkt, line 236: Please avoid `var`

autoform-helpers.js
```diff
@@ -270,12 +279,13 @@ Template.registerHelper('afFieldNames', function autoFormFieldNames(options) {
  * afSelectOptionAtts
  */
 Template.registerHelper('afSelectOptionAtts', function afSelectOptionAtts() {
-  var atts = _.pick(this, 'value');
+  if (this.value === false) this.value = 'false'
+  var atts = 'value' in this ? { value: this.value } : {}
```
jankapunkt, line 283: Please avoid `var`

components/afFormGroup/afFormGroup.html
```diff
@@ -1,3 +1,3 @@
 <template name="afFormGroup">
   {{> Template.dynamic template=getTemplateName data=innerContext}}
-</template>
+</template>
```
jankapunkt, line 3: Please add a newline add the end of the file

components/afFormGroup/afFormGroup.js
```diff
@@ -1,11 +1,11 @@
...
 
 Template.afFormGroup.helpers({
-  getTemplateName: function () {
-    return AutoForm.getTemplateName('afFormGroup', this.template, this.name);
+  getTemplateName: function() {
+    return AutoForm.getTemplateName("afFormGroup", this.template, this.name);
   },
   innerContext: function afFormGroupContext() {
-    var c = AutoForm.Utility.getComponentContext(this, 'afFormGroup');
+    var c = AutoForm.Utility.getComponentContext(this, "afFormGroup");
```
jankapunkt, line 8: Can we rename `c` to `context` or `ctx`?
Please avoid `var`.

components/afFormGroup/afFormGroup.js
```diff
@@ -15,30 +15,39 @@ Template.afFormGroup.helpers({
...
     // included multiple times, it's best not to provide an `id`
     // and generate a random one here for accessibility reasons.
-    var id = c.atts.id || Random.id();
-    var idPrefix = c.atts['id-prefix'];
-    if (idPrefix && idPrefix.length > 0) {
-      id = idPrefix + '-' + id;
+    const instance = Template.instance();
+    instance.fieldIds = instance.fieldIds || {};
+    const name = c.atts["name"];
+    var id = instance.fieldIds[name];
```
jankapunkt, line 21: Please avoid `var`

components/afFormGroup/afFormGroup.js
```diff
@@ -15,30 +15,39 @@ Template.afFormGroup.helpers({
...
+    if (!id) {
+      id = Random.id();
+      var idPrefix = c.atts["id-prefix"];
+      if (idPrefix && idPrefix.length > 0) {
+        id = idPrefix + "-" + id;
```
jankapunkt, line 26: Ca we use template literals?

components/afFormGroup/afFormGroup.js
```diff
@@ -72,8 +81,12 @@ function formGroupInputAtts(atts) {
   // Separate input options from label and formgroup options
   // We also don't want the 'label' option
   var inputAtts = {};
-  _.each(atts, function autoFormLabelAttsEach(val, key) {
-    if (['id-prefix', 'id', 'label'].indexOf(key) === -1 && key.indexOf('label-') !== 0 && key.indexOf('formgroup-') !== 0) {
+  Object.entries(atts).forEach(function autoFormLabelAttsEach([key, val]) {
+    if (
+      ["id-prefix", "id", "label"].indexOf(key) === -1 &&
+      key.indexOf("label-") !== 0 &&
+      key.indexOf("formgroup-") !== 0
```
jankapunkt, line 88: Can we use `includes` instead of `indexOf`?

components/autoForm/autoForm.js
```diff
@@ -1,12 +1,15 @@
...
 
 Template.autoForm.helpers({
   atts: function autoFormTplAtts() {
     // After removing all of the props we know about, everything else should
     // become a form attribute unless it's an array or object.
-    var val, htmlAttributes = {}, context = this;
+    var val,
+      htmlAttributes = {},
+      context = this;
     var removeProps = [
```
jankapunkt, line 13: Avoid `var`

components/autoForm/autoForm.js
```diff
@@ -26,19 +30,21 @@ Template.autoForm.helpers({
 
     // Filter out any attributes that have a component prefix
     function hasComponentPrefix(prop) {
-      return _.any(Utility.componentTypeList, function (componentType) {
-        return prop.indexOf(componentType + '-') === 0;
+      return Utility.componentTypeList.some(function(componentType) {
+        return prop.indexOf(componentType + "-") === 0;
```
jankapunkt, line 34: Prefer `includes`

components/autoForm/autoForm.js
```diff
@@ -95,44 +104,58 @@ Template.autoForm.created = function autoFormCreated() {
...
-      _.each(Hooks.getHooks(formId, 'docToForm'), function autoFormEachDocToForm(hook) {
-        doc = hook.call(hookCtx, doc, data._resolvedSchema);
-        if (!doc) {
-          throw new Error('Oops! Did you forget to return the modified document from your docToForm hook for the ' + formId + ' form?');
+      Hooks.getHooks(formId, "docToForm").forEach(
+        function autoFormEachDocToForm(hook) {
+          doc = hook.call(hookCtx, doc, data._resolvedSchema);
+          if (!doc) {
+            throw new Error(
+              "Oops! Did you forget to return the modified document from your docToForm hook for the " +
```
jankapunkt, line 138: Can we use template literals?

components/quickForm/quickForm.js
```diff
@@ -110,15 +107,15 @@ function getSortedFieldGroupNames(schemaObj) {
...
+  var fields = Object.entries(schemaObj).map(function ([fieldName, field]) {
     return (fieldName.slice(-2) !== '.$') &&
       field.autoform &&
       field.autoform.group === groupName &&
       fieldName;
   });
 
   // Remove undefined
-  fields = _.compact(fields);
+  fields = fields.filter(f => ![null, undefined, ''].includes(f));
```
jankapunkt, line 118: We can declare `f => ![null, undefined, ''].includes(f)` outside of the functions as it's re-used a lot in more than one method:

```
const emptyVals = [null, undefined, '']
const emptyFields = field => !emptyVals.includes(field)
```

and later use it like

```
fields = fields.filter(emptyFields);
```

formTypes/update-pushArray.js
```diff
@@ -21,14 +21,14 @@ AutoForm.addFormType('update-pushArray', {
...
+      if (!Object.keys(doc).length) { // make sure this check stays after the before hooks
         // Nothing to update. Just treat it as a successful update.
         c.result(null, 0);
       } else {
-        var modifer = {$push: {}};
+        var modifer = { $push: {} };
         modifer.$push[scope] = doc;
         // Perform update
-        collection.update({_id: c.docId}, modifer, c.validationOptions, c.result);
+        collection.update({ _id: c.docId }, modifer, c.validationOptions, c.result);
```
jankapunkt, line 31: Can we rename `c`?

formTypes/normal.js
```diff
@@ -34,15 +34,16 @@ AutoForm.addFormType('normal', {
...
+    }
 
     // Call all hooks at once.
     // Pass both types of doc plus the doc attached to the form.
     // If any return false, we stop normal submission, but we don't
     // run onError, onSuccess, endSubmit hooks until they all call this.done().
     var shouldStop = false;
-    _.each(hooks, function eachOnSubmit(hook) {
+    hooks.forEach(function eachOnSubmit(hook) {
       var result = hook.call(ctx, c.insertDoc, c.updateDoc, c.currentDoc);
```
jankapunkt, line 47: Can we rename `c`?

formTypes/update.js
```diff
@@ -15,12 +15,12 @@ AutoForm.addFormType('update', {
...
     // Run "before.update" hooks
     this.runBeforeHooks(this.updateDoc, function (modifier) {
-      if (_.isEmpty(modifier)) { // make sure this check stays after the before hooks
+      if (!Object.keys(modifier).length) { // make sure this check stays after the before hooks
         // Nothing to update. Just treat it as a successful update.
         c.result(null, 0);
       } else {
         // Perform update
-        collection.update({_id: c.docId}, modifier, c.validationOptions, c.result);
+        collection.update({ _id: c.docId }, modifier, c.validationOptions, c.result);
```
jankapunkt, line 23: Can we rename `c`?

versions.json
```diff
@@ -107,13 +107,9 @@
...
-    [
-      "underscore",
-      "1.0.1"
     ]
   ],
   "pluginDependencies": [],
   "toolVersion": "meteor-tool@1.0.35",
   "format": "1.0"
-}
+}
```
jankapunkt, line 115: Please add a newline at the end

utility.js
```diff
@@ -15,15 +16,18 @@ Utility = {
...
+      if (!Array.isArray(val) && isBasicObject(val)) {
         val = cleanNulls(val, false, keepEmptyStrings); // recurse into plain objects
-        if (!_.isEmpty(val)) {
+        if (Object.keys(val).length) {
           newDoc[key] = val;
         }
-      } else if (_.isArray(val)) {
+      } else if (Array.isArray(val)) {
+        if (!keepEmptyStrings) {
+          val = val.filter(v => ![null, undefined, ''].includes(v));
```
jankapunkt, line 27: Oh we can even make `v => ![null, undefined, ''].includes(v)` a common function as it's even used accross modules.

utility.js
```diff
@@ -124,7 +128,7 @@ Utility = {
 
     // Handle options="allowed"
     if (selectOptions === 'allowed') {
-      selectOptions = _.map(defs.allowedValues, function(v) {
+      selectOptions = defs.allowedValues.map(function (v) {
         var label = v;
         if (hash.capitalize && v.length > 0 && schemaType === String) {
           label = v.charAt(0).toUpperCase() + v.slice(1).toLowerCase();
```
jankapunkt, line 134: Can we use template literals?

utility.js
```diff
@@ -205,13 +209,13 @@ Utility = {
    */
   expandObj: function expandObj(doc) {
     var newDoc = {}, subkeys, subkey, subkeylen, nextPiece, current;
-    _.each(doc, function(val, key) {
+    Object.entries(doc).forEach(function ([key, val]) {
       subkeys = key.split('.');
       subkeylen = subkeys.length;
       current = newDoc;
       for (var i = 0; i < subkeylen; i++) {
```
jankapunkt, line 216: Since for loop with `break` are basically abstracted `goto` we should use a `forEach` alternative with `return`:

```javascript
const loop = [];
const loop.length = subkeylen;
loop.forEach((subkey, index) => {
  if (typeof current[subkey] !== 'undefined' && !isObject(current[subkey])) {
    return; // already set for some reason; leave it alone
  }

  if (index === subkeylen - 1) {
    ...
  }
})
```

utility.js
```diff
@@ -241,14 +245,12 @@ Utility = {
...
-        if (_.isArray(val)) {
-          obj[key] = _.without(val, void 0, null);
-          _.each(obj[key], function (arrayItem) {
-            compactArrays(arrayItem);
-          });
-        } else if (!(val instanceof Date) && _.isObject(val)) {
+    if (isObject(obj)) {
+      Object.entries(obj).forEach(function ([key, val]) {
+        if (Array.isArray(val)) {
+          obj[key] = val.filter(item => ![void 0, null].includes(item));
```
jankapunkt, line 251: `item => ![void 0, null].includes(item)` could be extracted into a module-level function

utility.js
```diff
@@ -241,14 +245,12 @@ Utility = {
...
+          obj[key].forEach(compactArrays);
+        } else if (!(val instanceof Date) && isObject(val)) {
```
jankapunkt, line 253: This is an example, where `isObject` is confusing - since it returns true also if val is an array or a function I would not know, whether the if condition will at this point be true because val is an object, an array or a function.

utility.js
```diff
@@ -475,11 +479,11 @@ Utility = {
 // getPrototypeOf polyfill
 if (typeof Object.getPrototypeOf !== 'function') {
```
jankapunkt, line 480: Is this obselete due to `web.legacy`?

utility.js
```diff
@@ -492,6 +496,6 @@ if (typeof Object.getPrototypeOf !== 'function') {
  * @param {any} obj
  * @returns {Boolean}
  */
-var isBasicObject = function(obj) {
-  return _.isObject(obj) && Object.getPrototypeOf(obj) === Object.prototype;
+var isBasicObject = function (obj) {
```
jankapunkt, line 499: Please avoid `var`

autoform-events.js
```diff
@@ -44,8 +50,8 @@ function endSubmit(formId, template, hookContext) {
 
 function adjustKeyForArrays(key) {
   var gKey = AutoForm.Utility.makeKeyGeneric(key);
-  if (gKey.slice(-2) === '.$' || gKey.indexOf('.$.') !== -1) {
-    key = gKey.slice(0, gKey.indexOf('.$'));
+  if (gKey.slice(-2) === ".$" || gKey.indexOf(".$.") !== -1) {
```
jankapunkt, line 53: Use `includes`

autoform-events.js
```diff
@@ -44,8 +50,8 @@ function endSubmit(formId, template, hookContext) {
...
+    key = gKey.slice(0, gKey.indexOf(".$"));
```
jankapunkt, line 54: Can we use template literals here?

autoform-events.js
```diff
@@ -312,20 +335,21 @@ Template.autoForm.events({
...
+      ...hookContext
+    });
   },
-  'keyup [data-schema-key]': function autoFormKeyUpHandler(event) {
+  "keyup [data-schema-key]": function autoFormKeyUpHandler(event) {
     // Ignore enter/return, shift, ctrl, cmd, tab, arrows, etc.
     // Most of these are just optimizations, but without ignoring Enter, errors can fail to show up
     // because of conflicts between running onSubmit handlers and this around the same time.
-    if (_.contains([13, 9, 16, 20, 17, 91, 37, 38, 39, 40], event.keyCode)) return;
+    if ([13, 9, 16, 20, 17, 91, 37, 38, 39, 40].includes(event.keyCode)) return;
```
jankapunkt, line 352: This array can be declared as const on the module level

autoform-inputs.js
```diff
@@ -2,11 +2,12 @@
 
 getFlatDocOfFieldValues = function getFlatDocOfFieldValues(fields, ss) {
   var doc = {};
-  fields.each(function () {
-    var fieldName, val = AutoForm.getInputValue(this, ss);
+  fields.each(function() {
+    var fieldName,
```
jankapunkt, line 6: Please avoid `var`

autoform-inputs.js
```diff
@@ -148,83 +163,110 @@ getInputData = function getInputData(defs, hash, value, label, formType) {
...
 
-updateTrackedFieldValue = function updateTrackedFieldValue(template, fieldName) {
-  if (!template) return;
+const markChangedAncestors = (template, fieldName) => {
+  // To properly handle array fields, we'll mark the ancestors as changed, too
+  // FIX THIS
+  // XXX Might be a more elegant way to handle this
 
-  template.formValues = template.formValues || {};
+  var dotPos = fieldName.lastIndexOf(".");
```
jankapunkt, line 193: Please avoid `var`

components/afFieldInput/afFieldInput.js
```diff
@@ -1,5 +1,78 @@
...
+  const instance = Template.instance();
+  updateTrackedFieldValue(
+    template,
+    instance.afFieldName,
+    instance.afFieldValue
+  );
+});
+
+export function afFieldInputContext() {
+  var c = AutoForm.Utility.getComponentContext(this, "afFieldInput");
```
jankapunkt, line 14: Please avoid var and rename `c` to something more verbose

components/afFieldInput/afFieldInput.js
```diff
@@ -1,5 +1,78 @@
...
+  const template = AutoForm.templateInstanceForForm();
+  const instance = Template.instance();
+  updateTrackedFieldValue(
+    template,
+    instance.afFieldName,
+    instance.afFieldValue
+  );
+});
+
+export function afFieldInputContext() {
```
jankapunkt, line 13: Please comment on the function with a short summary

components/afFieldInput/afFieldInput.js
```diff
@@ -10,19 +83,28 @@ Template.afFieldInput.helpers({
...
+      inputTemplateName,
+      styleTemplateName,
+      self.name,
+      true
+    );
 
     // Special case: the built-in "plain" template uses the basic input templates for
     // everything, so if we found _plain, we use inputTemplateName instead
-    if (templateName.indexOf('_plain') !== -1) {
+    if (templateName.indexOf("_plain") !== -1) {
```
jankapunkt, line 107: Please use `includes`

formTypes/method-update.js
```diff
@@ -1,56 +1,77 @@
 /* global AutoForm */
 
-AutoForm.addFormType('method-update', {
-  onSubmit: function () {
+AutoForm.addFormType("method-update", {
+  onSubmit: function() {
     var c = this;
```
jankapunkt, line 5: Please avoid `var` and rename `c` to something understandable

formTypes/method-update.js
```diff
@@ -1,56 +1,77 @@
...
       // Validate. If both schema and collection were provided, then we validate
       // against the collection schema here. Otherwise we validate against whichever
       // one was passed.
-      var valid = (c.formAttributes.validation === 'none') ||
-          c.formTypeDefinition.validateForm.call({
-            form: c.formAttributes,
-            formDoc: updateDoc,
-            useCollectionSchema: c.ssIsOverride
-          });
+      var valid =
```
jankapunkt, line 21: Please avoid `var`

formTypes/method-update.js
```diff
@@ -1,56 +1,77 @@
...
       } else {
+        const { methodargs } = c.formAttributes;
+        const args = methodargs
+          ? typeof methodargs === "function"
+            ? methodargs()
+            : methodargs
+          : [];
         // Call the method. If a ddp connection was provided, use
         // that instead of the default Meteor connection
         var ddp = c.formAttributes.ddp;
```
jankapunkt, line 40: Please avoid `var`

formTypes/method-update.js
```diff
@@ -1,56 +1,77 @@
...
+          c.result
+        );
       }
     });
   },
   usesModifier: true,
-  validateForm: function () {
+  validateForm: function() {
     // Get SimpleSchema
     var ss = AutoForm.getFormSchema(this.form.id);
```
jankapunkt, line 57: Please avoid `var`

formTypes/method.js
```diff
@@ -1,52 +1,68 @@
 /* global AutoForm */
 
-AutoForm.addFormType('method', {
-  onSubmit: function () {
+AutoForm.addFormType("method", {
+  onSubmit: function() {
     var c = this;
```
jankapunkt, line 5: Please avoid `var` and rename `c`

formTypes/method.js
```diff
@@ -1,52 +1,68 @@
...
       // Validate. If both schema and collection were provided, then we validate
       // against the collection schema here. Otherwise we validate against whichever
       // one was passed.
-      var valid = (c.formAttributes.validation === 'none') ||
-          c.formTypeDefinition.validateForm.call({
-            form: c.formAttributes,
-            formDoc: doc,
-            useCollectionSchema: c.ssIsOverride
-          });
+      var valid =
```
jankapunkt, line 21: Please avoid `var`

formTypes/method.js
```diff
@@ -1,52 +1,68 @@
...
       } else {
+        const { methodargs } = c.formAttributes;
+        const args = methodargs
+          ? typeof methodargs === "function"
+            ? methodargs()
+            : methodargs
+          : [];
         // Call the method. If a ddp connection was provided, use
         // that instead of the default Meteor connection
         var ddp = c.formAttributes.ddp;
```
jankapunkt, line 40: Please avoid `var`

formTypes/method.js
```diff
@@ -1,52 +1,68 @@
...
+        ddp.call(c.formAttributes.meteormethod, doc, ...args, c.result);
       }
     });
   },
-  validateForm: function () {
+  validateForm: function() {
     // Get SimpleSchema
     var ss = AutoForm.getFormSchema(this.form.id);
 
     var collection = AutoForm.getFormCollection(this.form.id);
```
jankapunkt, line 50: Please avoid `var`

jankapunkt (changes requested):
I finally managed to make a full review of the code. There are minor issues but they are spread accross the code. I would mostly care about the following

**code**

- `isObject` has a truthy return value also for arrays and functions which makes it hard to debug any condition because we never know, whether `isObject` was true, due to a value being an object, array or function. It caused no runtime errors for me but future debugging sessions could be more exhausting due to this unknown state
- avoid `var` wherever possible and use `const` for all non-redeclarable variables, otherwise `let`
- use template literals instead of manually concatenating Strings
- replace `indexOf(...) > -1` and similar with `.includes` where the index itself is not of interest (but not with `lastIndexOf` which can't be replaces by `includes`)
- Rename single-char variables to something more understandable
- sometimes the multiline formatting (especially function paramters) can make it harder to read but this is maybe also a personal thing of me


**documenation**

- Add some short summary on undocumented methods of any class / higher Object
- Make notes on new or altered behavior
- Make notes (if possible) if new / altered behavior fixes a related issue
- Update `API.md`

**consequences of changes** 

The most changes may not have consequences, however

- if a `var` declared variable was beforehand not block-scoped and the code unintentionally assumed this variable to be global or not block-scoped this could lead to an error. However it is then not the fault of the variable being declared as `const` or `let` but because of wrong scoping and the scoping should be fixed, rather than leaving the variable as `var`
- the `isObject` currently represents also a truthy state for array and function values but should not do this. instead each case should tested separately. This may cause errors on conditions, that implicitly expect a value to bei either an object, array or function (and where `isObject` is currently in correct use).

Both should be handled with care and tested thoroughly.

pouya-eghbali:
The `var`,`indexOf` and single char / short names are not my doings, I tried to use `const` whenever I rewrote any part of this, and if that wasn't possible I used `let`, I tried my best not to touch the original code but if we're doing a major update we should replace all these and move to >es6.

For the breaking of parameters and formatting, I used [prettier](https://prettier.io) with default settings, we can decide on a linter and add a config file to automatically format the code. I used prettier with default settings because that's what I use for all my projects.

All changes I made are somehow documented in the new readme, this was fine for a fork, but we need to move it to the documentation. Regarding the `isObject` function, do you know a standalone fast implementation?

pouya-eghbali:
For `isObject` the current implementation is

```javascript
export const isObject = obj => toString.call(obj) == '[object Object]' || Array.isArray(obj) || isFunction(obj)
```

can't remember the reason behind this, we should check if removing ` || Array.isArray(obj) || isFunction(obj)` won't break any functionality, or if it does then it should be replaced with an appropriate function call where it breaks.

jankapunkt:
> The var,indexOf and single char / short names are not my doings, I tried to use const whenever I rewrote any part of this, and if that wasn't possible I used let, I tried my best not to touch the original code but if we're doing a major update we should replace all these and move to >es6.
>
> For the breaking of parameters and formatting, I used prettier with default settings, we can decide on a linter and add a config file to automatically format the code. I used prettier with default settings because that's what I use for all my projects.

I see, I was already curious because at some parts there wire mixed `var` and `const`. I think prettier is a stable and well established tool and we could go for it. I could setup some github action / ci using it to lint-test any pushes and pull requests (will create a PR for that).

If you give me access to the dev repo I can take over that part with the ES6 and fix all my mentioned things in the review. An alternative would be to merge this into a `v7.x` branch instead of `devel` and work from here towards a stable v7. What do you think?

> can't remember the reason behind this, we should check if removing || Array.isArray(obj) || isFunction(obj) won't break any functionality, or if it does then it should be replaced with an appropriate function call where it breaks.

I think the function should not be too complicated and fulfill the following requirements:

- check if it's not an Array
- check if it's not a Function

I checked the [polyfill for `Array.isArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Polyfill) and would do

```javascript
export const isObject = obj => Object.prototype.toString.call(obj) === '[object Object]'
```

which is basically the explicit version of `toString.call(obj) == '[object Object]'` and in any situation it breaks there should be an extra condition with `else if (Array.isArray(obj)) { ... }` and/or `else if (typeof obj === 'function') { ... }```

jankapunkt:
I just want to ask again, do you think it's a good idea to merge this to a `v7` branch instead of `develop` so I can immediately fix the issues that are left and we can soon merge? When I am through I will ask for at least one reviewer from the team and then we should soon be able to merge.

pouya-eghbali:
Yeah, let's do that. That sounds like a better idea.

jankapunkt:
Allright, changed the base branch and will implement the mentioned issues, I hope this will be done this weekend :-)

PR #1694, opened 2020-06-18
v7.0.0

### LINKED ISSUE #1692: Upgrade to 7.0.0

### LINKED ISSUE #1696: afQuickField id field is overwritten - v7
This issue was introduced in 7, working previously.

We have a modal with a datetimepicker in it (eonasdan-bootstrap-datetimepicker) which is hidden. Before showing the modal we try and set the date in the picker but the picker has not been initialised so fails. 

The template that the modal contains has been rendered so I can only imagine that there might be an optimisation to not init controls that are not visible somewhere in 7?

jankapunkt:
This PR finalizes the rewrite from #1693 on this repo's `v7-dev` branch. This includes proper update of the mentioned issues from #1693, a solid testing and a great and updated documentation.

## What's the goal with v7?

This version improves performance a lot, while at the same time reducing bundle size. No underscore anymore and the (outdated BS3) templates are now decoupled from the package, so you are highly flexible in choosing a theme template for your forms.

Also the list of extensions is curated to highlight deprecated/abandoned packages.

## Why not finalize in #1693 ?

The #1693 was done by @pouya-eghbali (great job!) and come from his fork. In order to allow multiple contributions on v7 without the need of managing a second repo, we just merged it into `v7-dev` and continue to work on this until we have a final version.

## TODO List from #1693 review

- [x] replace all `var` with ES6 `let` or `const` and test of the scoping does not break anything
- [x] rename unclear variables, like `c` to have meaningful names
- [x] replace `indexOf` and -1 comparison on Arrays with `includes`
- [x] `isObject` -> `export const isObject = obj => Object.prototype.toString.call(obj) === '[object Object]'`
- [x] check for double/multiple declared functions and merge them into a single export
- [x] Testing: update testing suite to `meteortesting:mocha` v2
- [x] Testing: introduce GitHub actions to make CI maintenance easier
- [x]  allow dynamic imports via process.env flag without break actual functionality
- [x] the API docs need to be rebuild, since there are updated / changed jsDoc comments in the code

Additional bugs to fix, introduced with this PR:

- [x] #1696

## Post-release things to do

- Discussion: make `moment` a weak dependency, replace it with luxon / day.s or even with native Date?
- The documentation needs to be updated (review outdated extensions, remove links to the old demo)
- The demo has been shut down, so we need a good usable demo (using npm `meteor-build-client`) to publish in GitHub pages

## How can I help?

We need reviewers and testers as much as possible to make this a stable version. Breaking changes and breaking compatibility with extensions need to be documented, as well as newly added functions. Tests should be added where possible.


package.js
```diff
@@ -1,241 +1,174 @@
...
+      "deps",
+      "templating",
+      "ui",
+      "blaze",
+      "ejson",
+      "reactive-var",
+      "reactive-dict",
+      "random",
+      "ecmascript",
+      "momentjs:moment@2.10.6"
```
jankapunkt: @pouya-eghbali I checked with bundle visualizer and found 60kb due to the momentjs package being a hard dependency. Is there a reason against putting it into the weak dependencies?

package.js
```diff
@@ -1,241 +1,174 @@
...
-    'templates/plain/components/quickForm/quickForm.js',
-    'templates/plain/components/afArrayField/afArrayField.html',
-    'templates/plain/components/afFormGroup/afFormGroup.html',
-    'templates/plain/components/afObjectField/afObjectField.html',
-    'templates/plain/components/afObjectField/afObjectField.js',
-    // plain-fieldset Template
-    'templates/plain-fieldset/plain-fieldset.html',
-    'templates/plain-fieldset/plain-fieldset.js',
-  ], 'client');
+  api.addFiles(
```
jankapunkt: @pouya-eghbali what do you think about removing these and instead provide a dynamic import via `AutoForm.initialize()` function? That would reduce initial bundle sizes by nearly 80kb. We could also keep static imports, based on a process.env flag, like

```javascript
if (process.env.AUTOFORM_STATIC_IMPORT) {
  api.addFiles([
    ...
  ]);
}
```

jankapunkt:
@pouya-eghbali  @Meteor-Community-Packages/autoforms  while working on this PR I came across several issues that I'd like to discuss with you all.

## Make local testing easy and accessible

The `api.exports` and `api.addFiles` makes it impossible to inject `mongo-object` NPM package on a local package test without creating a host Meteor project for tests.

Improving tests would be an important part of maintaining this package but in the current structure this would be a real hazzle.

Also - CI testing could be improved a lot when we change the package exports signature. I would introduce GitHub actions as a second automated testing besides Travis

Plus - Global namespace would have one entry less :tada:

Disadvantage - extensions would have to be updated in order to actively import AutoForm in order to make use of it.

What I would propose - change the package structure to use an `api.mainModule` and change tests to use `meteortesting:mocha` with `chai`

## Introduce dynamic imports

Currently this package adds 80kb to the inital bundle size (plus extensions that are also loaded immediately) allthough I seldom load a form immediately (besides logins)

The above proposed `api.mainModule` structure could also allow the introduction of (optional) dynamic imports, that can be controlled using a `process.env` flag.

## MomentJS

If we get the above dynamic import structure working, we could also introduce dynamic loading of moment from the npm package, which would again reduce the initial bundle  by 60KB

What do you people thing?

jankapunkt:
# Big Improvements here

Okay this is getting interesting. By the time I was trying to move ES5 to ES6 code I was so unsatisfied with the fact, that I don't know if this breaks anything. I started writing tests until the point where we now have a comprehensive test suite for the API and the internals it uses (ArrayTracker, FormData, Hooks, Inputs, Validation etc.).

The tests are stubbing out external behavior (SimpleSchema, MongoObject), because that would be impossible for one person to implement with limited time and resources. However the current tests are quite solid (imo), so let me summarize what we have up to this point:

- The package is easily testable now (see the updated sections in the new README)
- 229 Unit tests, covering API and internals
- GitHub CI running on push and PR, for now only on v7-dev until the branch is merged into develop
- API and the internals it uses are rewritten from ES5 to ES6
- To make the internals more testable there were some exports added
- I tried to not change any logic so there should nothing be broken (also tested in my current project that uses this form)
- Found 1-2 bugs that have been revealed during the test writing that have been fixed + commented

It would be more than awesome if someone would take on a first pre-review and check, if everything works fine. There are also still some points left (see initial post of this PR) and there is still some work for prettifier to be done on the files and coverage would also be nice2have. However, I first wanted to have some stable tests in order to see if improvements won't break anything. 

I will continue to work on the points left. In the meantime it would be awesome to have people to help with testing.

StorytellerCZ:
@jankapunkt Might be a good idea to do RC releases so it is easier more convenient for people to test and to get it on their radar (also announce on the forums).

jankapunkt:
Thank you @StorytellerCZ I would go forward with this as a `7.0.0-rc` and will then make an announcement. However I am not allowed to publish here, so I will still need someone with publish permissions to be ready at that point.

StorytellerCZ:
@jankapunkt I could add you or I can do it.

jankapunkt:
This package is now ready to be published, I am just waiting to get the themes to be published first and then we can finally make it official.

PR #1695, opened 2020-07-03
Label icons for accessibility

emollett:
The buttons to add or remove items have no accessible text for screen reader users.

aria-label added to add and remove buttons for the afArrayField for accessibility

### DIFF
--- a/templates/bootstrap3-horizontal/components/afArrayField/afArrayField.html
+++ b/templates/bootstrap3-horizontal/components/afArrayField/afArrayField.html
@@ -14,7 +14,7 @@
             <div>
               <div class="autoform-remove-item-wrap">
                 {{#if afArrayFieldHasMoreThanMinimum name=../atts.name minCount=../atts.minCount maxCount=../atts.maxCount}}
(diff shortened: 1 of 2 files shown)


jankapunkt:
Hi @emollett 

thank you for your contribution. Improving accessibility is always a good idea. However, we may have to consider that this component is used by non-English speakers, too. Thus, it needs to be configurable to let the developer define what word to display.

Please take a look at how the submit button label is defined in a `quickForm`. By default it is defined as `Submit` but you can easily override it using `button-content="Whatever submit text". I think this is the right way to go.

emollett:
@jankapunkt thanks that makes sense. Thanks for the pointer on overriding the button content, my understanding had been this is just for the submit button and not the add/remove ones. I understand I can make my own template for these though. For now as a quick fix I'm adding content to the buttons with css which works for the accessibility needs in this instance.

jankapunkt:
Hey @emollett no need to close the pr. You just add your changes as new commits to your fork and we squash-merge it all at the end into one commit.

I think you don't need a full template, just a helper and a reactive variable, as the submit button does.

PR #1701, opened 2020-12-29
Meteor AutoForm 7.1.0 - Autocomplete

### LINKED ISSUE #1700: Create a native 'autocomplete' type for `afQuickfield` in `AutoForm` that is Bootstrap themed.
## Problem
In 2020 and beyond, `autocomplete` input fields are a common pattern for data-intensive form completion, but there is no native solution in the AutoForm boilerplate. Many of today's applications require "rich" input experience that can quickly provide suggestions filtered from already existing entries in the database, or an enumerable list of options. Using a "select" for everything is very cumbersome, especially as the list of possible solutions to the text matching enumerable entries exceeds the available area of the UI viewport, such as on Mobile devices. The existing `autocomplete` and `typeahead` "add-ons" are outdated and don't conform to the v7 API design, pull in external dependencies, and don't support Bootstrap 4 and beyond.

## Solution
Generally, rather than rely exclusively on select dropdown elements, a proper experience leverages a "text" input field that offers a list of autocomplete solutions matching the input text (the needle) with a datum (the haystack) using one or more user-defined patterns (e.g. start, global, case sensitive, case insensitive, all via RegEx).

Specifically, we need to create the `/inputTypes/autocomplete/` directory that contains `autocomplete.js` and `autocomplete.html`. The basic code for these is included below.

    // autocomplete.js
    AutoForm.addInputType('autocomplete', {
      template: 'afAutocomplete',
      valueOut: function () {
        return this.val()
      },
      valueConverters: {
        stringArray: AutoForm.valueConverters.stringToStringArray,
        number: AutoForm.valueConverters.stringToNumber,
        numberArray: AutoForm.valueConverters.stringToNumberArray,
        boolean: AutoForm.valueConverters.stringToBoolean,
        booleanArray: AutoForm.valueConverters.stringToBooleanArray,
        date: AutoForm.valueConverters.stringToDate,
        dateArray: AutoForm.valueConverters.stringToDateArray
      },
      contextAdjust: function (context) {
        context.atts.autocomplete = 'off'
        return context
      }
    })

    // autocomplete.html
    <template name="afAutocomplete">
      <input type="text" value="{{this.value}}" {{this.atts}} />
    </template>

Moreover, the `autocomplete.js` should include helpers and UI code that populates Mtext inside the input field plus a floating list of elements below or above the text input field, which the user can highlight using arrow keys and select using `Return`, touch, or mouse click. 

Options for the autocomplete can be set two ways: 

1. Via static `autoform.options` inside a `simpl-schema` definition. Much like the options used in a `select` type.
2. Via `ReactiveVar` or function helper set on `data-autocomplete-options` in the user's `afQuickfield` definition. This offers the power to tie a capped collection, or optimized query from MongoDB into the UX, allowing for a dynamic autocomplete experience.

CSS for the autocomplete field should come from Bootstrap 4 (or later) and implement a structured set of class selector definitions for the elements that make up the autocomplete box, which could be structured as follows:

Lastly, we need to allow for theme implementations by adding a Bootstrap 4 theme in `meteor-autoform-themes` in `/bootstrap4/templates/bootstrap4/inputTypes/` so that the `form-control` class and styles get applied. 

    <template name="afAutocomplete_bootstrap4">
      <input type="hidden" value="{{this.label}}" {{ this.visibleAtts }} />
      <input type="text" value="{{this.value}}" {{ this.visibleAtts }} />
      <div class="autocomplete dropdown">
        <div class="ac-option">First</div>
        <div class="ac-option ac-hilite">Second</div>
      </div>
    </template>

With as common and data-enriched applications are becoming, this feature could also revive a little more popularity for the Meteor framework and AutoForm by adding an elegant solution to the quiver of modern application and UX development tools.

## Alternatives Considered
All the suggested add-on packages were taken into consideration and evaluated prior to deciding to pursue this. None are up-to-date and all rely on external libraries for functionality, such as the old `typeahead` solution from Bootstrap 3 (now deprecated). The use of a `select` was also taken into consideration, but as mentioned, proved inefficient. In the use case that we are implementing in our venture, we require a live list of currency "asset pairs" that varies between exchanges, so we demand a reactive list of autocomplete options to speed the process of populating Orders.

## Additional context
This issue is the product of [a conversation started on Stack Overflow](https://stackoverflow.com/questions/65298170/where-does-the-form-control-class-get-set-when-using-autoform-bootstrap-4) relating to issues associated with wanting to implement a custom `autocomplete` for AutoForm.

I'm going to move forward with this in the next few days for our project. I'd be humbled to contribute this back to the community if others find this useful and appropriate.

4Z4T4R:
- Added new inputTypes autocomplete directory
- Added HTML for Autocomplete Template
- Added JS for Autocomplete Template
- Circular selection allows arrows to cycle tip to tail and vice versa
- Reactive options allow Tracker vars to provide labels and values to dropdown selections
- Simpl-Schema defined options provide the same path to values and labels as select lists.
- Navigable via arrows so your mouse can rest.
- No third party solutions so we're not feeling bloated.
- Uses bootstrap 4 dropdown class definitions for style so adopters can use theme packages, no problemo.
- Conforms to StandardJS delinting and format so that the gods (and developers) are happy.

### DIFF
--- a/inputTypes/autocomplete/autocomplete.js
+++ b/inputTypes/autocomplete/autocomplete.js
@@ -17,16 +17,17 @@ AutoForm.addInputType('autocomplete', {
   },
   contextAdjust: function (context) {
     context.atts.autocomplete = 'off'
-    const itemAtts = { ...context.atts }
+    const { ...itemAtts } = context.atts
     // remove non-essential atts from visible input
-    const visibleAtts = Object.assign({}, { ...context.atts })
-    const keys = ['data-schema-key', 'id', 'name']
-    keys.forEach(key => {
+    const visibleAtts = Object.assign({}, context.atts)
+
+    ;['data-schema-key', 'id', 'name'].forEach(key => {
       delete visibleAtts[key]
     })
+
     // add form-control to remaining classes
-    context.visibleAtts = AutoForm.Utility.addClass({ ...visibleAtts }, 'form-control')
-    context.atts = AutoForm.Utility.addClass({ ...itemAtts }, 'form-control')
+    context.visibleAtts = visibleAtts
+
     // build items list
     context.items = []
 
@@ -121,6 +122,11 @@ Template.afAutocomplete.onRendered(function () {
     }
   })
 
+  const updateValue = value => {
+    $hidden.val(value)
+    $hidden.trigger('change')
+  }
+
   // clear on blur?
   // TODO: Figure out how blur won't block "click"
   // $input.blur((e)=>{ clearDropdown(e) })
@@ -131,7 +137,7 @@ Template.afAutocomplete.onRendered(function () {
     if (/ArrowDown|ArrowUp|ArrowLeft|ArrowRight|Enter|Escape/.test(e.originalEvent.key) === false) {
       // we're typing
       // ensure hidden and visible values match for validation
-      $hidden.val($input.val())
+      updateValue($input.val())
       // filter results from visible input value
       const result = items.get().filter((i) => {
         const reg = new RegExp(e.target.value, 'gi')
@@ -166,7 +172,7 @@ Template.afAutocomplete.onRendered(function () {
           const dataValue = me.$(e.target).attr('data-value')
           const dataLabel = me.$(e.target).attr('data-label')
           $input.val(dataLabel)
-          $hidden.val(dataValue)
+          updateValue(dataValue)
           clearDropdown(e, false)
           $input.focus()
         })
@@ -176,7 +182,7 @@ Template.afAutocomplete.onRendered(function () {
         // bc we all make mistakes
         if (result.length === 1) {
           $input.val(result[0].label)
-          $hidden.val(result[0].value)
+          updateValue(result[0].value)
           clearDropdown(e, false)
           $input.focus()
         }
@@ -214,12 +220,16 @@ Template.afAutocomplete.onRendered(function () {
         const enterValue = $suggestions.children('div').eq(currIndex).attr('data-value')
         const enterLabel = $suggestions.children('div').eq(currIndex).attr('data-label')
         $input.val(enterLabel)
-        $hidden.val(enterValue)
+        updateValue(enterValue)
         clearDropdown(e, false)
         $input.focus()
       }
     }
   }
+
+  $input.blur(() => {
+    $hidden.trigger('blur') // triggers re-validation
+  })
 
   // detect keystrokes
   $input.keyup((e) => {
@@ -233,6 +243,7 @@ Template.afAutocomplete.onRendered(function () {
 
   // show on double click
   $input.on('touchstart', (e) => {
+    $hidden.trigger('touchstart')
     callback(e)
   })
 })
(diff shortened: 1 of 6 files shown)


inputTypes/autocomplete/autocomplete.js
```diff
@@ -0,0 +1,167 @@
...
+
+        // clear any manual navigated selections on hover
+        $suggestions.children().hover((e) => {
+          $suggestions.children().removeClass('active')
+          currIndex = -1
+        })
+
+        // choose an answer on click
+        $suggestions.children().click((e) => {
+          const dataValue = $(e.target).attr('data-value')
```
jankapunkt: can you please change this to

```javascript
const dataValue = me.$(e.target).attr('data-value')
``` 

so it uses the local-scoped jQuery selector plus it resolves the linter error.

4Z4T4R: Done. Thx.

inputTypes/autocomplete/autocomplete.js
```diff
@@ -230,4 +230,9 @@ Template.afAutocomplete.onRendered(function () {
   $input.dblclick((e) => {
     callback(e)
   })
+
+  // show on double click
```
4Z4T4R, line 272: change comment to `show on touch`
Good catch!

inputTypes/autocomplete/autocomplete.js
```diff
@@ -121,6 +122,11 @@ Template.afAutocomplete.onRendered(function () {
     }
   })
 
+  const updateValue = value => {
+    $hidden.val(value)
+    $hidden.trigger('change')
+  }
+
```
4Z4T4R: Clever! Nicely done. Is it safe to assume that the change event triggers validation? on blur too?

inputTypes/autocomplete/autocomplete.js
```diff
@@ -214,13 +220,17 @@ Template.afAutocomplete.onRendered(function () {
...
         clearDropdown(e, false)
         $input.focus()
       }
     }
   }
 
+  $input.blur(() => {
+    $hidden.trigger('blur') // triggers re-validation
+  })
+
```
4Z4T4R: Ah! There it is. Well done. Thx!

inputTypes/autocomplete/autocomplete.js
4Z4T4R: I see the blur change below.

inputTypes/autocomplete/autocomplete.js
```diff
@@ -121,6 +122,11 @@ Template.afAutocomplete.onRendered(function () {
...
   })
 
+  const updateValue = value => {
+    $hidden.val(value)
+    $hidden.trigger('change')
+  }
+
   // clear on blur?
   // TODO: Figure out how blur won't block "click"
   // $input.blur((e)=>{ clearDropdown(e) })
```
4Z4T4R: Remove this comment.

4Z4T4R (commented):
There's general feedback here. Just some curiosity around the trigger for the input/hidden input.

inputTypes/autocomplete/autocomplete.js
jankapunkt: The change is used to fire Updates for any reactive getter like AutoForm.getFieldValue

While blur triggers validation.

PR #1708, opened 2021-05-11
add all test files to .meteorignore

jankapunkt:
I totally forgot to add tests and test-related folders to the `.meteorignore` which could cause trouble for some when deploying their apps

### DIFF
--- a/.meteorignore
+++ b/.meteorignore
@@ -1,0 +1,3 @@
+/testapp
+/tests
+**/*tests.js


StorytellerCZ (approved):

copleykj (approved):

.meteorignore
```diff
@@ -0,0 +1,3 @@
+/testapp
+/tests
```
copleykj, line 2: meteor seems to include the tests directory and it's contents even if it is included in `.meteorignore`, probably because it's a special directory.

PR #1734, opened 2023-08-22
Feature: warn on no theme

### LINKED ISSUE #1731: Please add error/warning messages if no theme is installed
Now that autoform 7 decouples themes, you must manually install and activate a theme separately. But if you don't the form simply doesn't render and there are no client or server errors.

Could you please add an error/warning message if there is no theme installed or activated?

jankapunkt:
Raises a warning, if no theme is installed. The warning itself is also covered in the README.

### DIFF
--- a/autoform-api.js
+++ b/autoform-api.js
@@ -309,6 +309,7 @@ AutoForm.getTemplateName = function autoFormGetTemplateName (
   }
 
   // Found nothing. Return undefined
+  console.warn('AutoForm: no theme found, please consult the README.')
 }
 
 /**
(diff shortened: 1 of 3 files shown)


autoform-api.js
```diff
@@ -309,6 +309,7 @@ AutoForm.getTemplateName = function autoFormGetTemplateName (
   }
 
   // Found nothing. Return undefined
+  console.warn('AutoForm: no theme found, please consult the README.')
```
StorytellerCZ, line 312: In the future we could use the `logging` package to do something like `Log.debug`, but since we need this warning before this comes anywhere near production this is adequate.

StorytellerCZ (approved):

StorytellerCZ:
@jankapunkt failing test, probably a bad selector judging from the error...

jankapunkt:
Weird, since I only added a warn output. 🤔

PR #1738, opened 2024-01-02
Add mongo-object

harryadel:
Autoform relies on mongo-object internally but doesn't state it as an explicit NPM dependency because it's one of the sub dependencies of simple-schema which you get it by default when you install simpl-schema. The problem comes apparent only when you remove the simpl-schema NPM package.

### DIFF
--- a/package.js
+++ b/package.js
@@ -7,6 +7,10 @@ Package.describe({
   version: '7.0.1'
 })
 
+Npm.depends({
+  'mongo-object': '3.0.1'
+});
+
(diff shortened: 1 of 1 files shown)


jankapunkt:
Shouldn't we rather api.mply Mongo Object in the simple schema Meteor package?

harryadel:
@jankapunkt what led me to create this issue is that once I removed the NPM version, I had a client error stopping my application from booting till I installed the mongo-object as an NPM version.

